PR #29, opened 2026-07-26
Let the Phase 0 workflow run FND-001b and FND-001c

afternoon:
## Summary

`FND-001b` and `FND-001c` were already defined in `docs/backlog.md`, but the Phase 0 workflow's `TASKS` array only listed `FND-001` through `FND-009` with no `b`/`c` variants — so **no run of the workflow could ever produce them**. `FND-009` depends on `FND-001c`, which depends on `FND-001b`, so Phase 0's exit gate had no path to its analytics and hosted-environment criteria.

This adds both to the workflow. It changes one file and does not touch the backlog — no task was split, renamed, or redefined.

## Changes

- **Both tasks added to the Tooling phase**, running strictly in order and never concurrently: the deploy pipeline must exist before analytics can stamp a release against it or upload source maps through it. Each gate fires only when the task actually ran, so a subset filter that omits one doesn't halt the run.

- **`FND-001c` joins `CONTRACT_TASKS`**, and so runs on Opus, because its own backlog block declares it contract-owning — it publishes the catalog every Phase 1-4 feature task extends.

- **Fixed a latent bug in the task-id filter.** `normaliseId` upper-cased the caller's ids but compared them against raw task ids. That was invisible while every id was already upper-case, but `FND-001b` normalises to `FND-001B`, so **every filter naming these two tasks would have been rejected as an unknown task id**. Both sides normalise now. Verified across all five accepted filter forms, and confirmed the malformed-filter guards (`''`, `'[]'`, `42`, unknown ids) still throw.

- **`NO_CREDENTIALS` brief** for `FND-001b`, `FND-001c`, and `FND-009` — the only Phase 0 tasks depending on things outside the repository (a Firebase project, a GA4 property, a Sentry org, CI secrets for all three). An agent cannot provision those, and the failure mode isn't stopping: it's inventing a plausible project id or DSN, committing it, and reporting a deploy that never happened. The brief forbids that and asks for the unverifiable checkboxes back as honest `unmet` entries.

- **`FND-009` gated on its dependencies having landed**, counting only tasks the same invocation attempted, so it isn't sent into two review rounds it cannot pass. Already-merged tasks aren't in `results` and are correctly not treated as missing.

## Verified in use

All three tasks were run through the workflow with these changes. `FND-001b` and `FND-007` are open as #25 and #26; `FND-001c` is #27, stacked on #25. The `NO_CREDENTIALS` brief did its job — each task returned honest `unmet` entries for the provisioning-dependent checkboxes rather than fabricating evidence.

Two limitations surfaced during that run and are **not** addressed here; see the review comment below.

https://claude.ai/code/session_01VMeruF1MYJavtosdz4bZhg

### DIFF
--- a/.claude/workflows/solid-groove-phase-0.js
+++ b/.claude/workflows/solid-groove-phase-0.js
@@ -5,7 +5,7 @@ export const meta = {
   whenToUse:
     'Run to execute Phase 0 of docs/backlog.md. Name tasks to run a subset, either positionally (solid-groove-phase-0 FND-003 FND-004) or as args: ["FND-003","FND-004"]. Omit them entirely to run the whole phase. Named tasks still execute in dependency order, not the order given.',
   phases: [
-    { title: 'Tooling', detail: 'FND-001 — test, CI and emulator foundation' },
+    { title: 'Tooling', detail: 'FND-001 test and CI foundation, then FND-001b deploy pipeline and FND-001c analytics catalog' },
     { title: 'Contracts', detail: 'FND-002 domain schema, then command kernel, repository and projections', model: 'opus' },
     { title: 'Runtime', detail: 'FND-006 AudioRuntime and FND-008 renderer harness' },
     { title: 'Graph', detail: 'FND-007 stable ID-keyed audio graph' },
@@ -17,7 +17,9 @@ export const meta = {
 // contract-owning tasks run on Opus because an error there propagates into every
 // dependent task and surfaces late; everything else runs on Sonnet. Review is
 // always Opus, at high effort, and runs before any PR is opened.
-const CONTRACT_TASKS = ['FND-002', 'FND-003', 'FND-004']
+// FND-001c is included because its own backlog block declares it contract-owning:
+// it publishes the analytics catalog every Phase 1-4 feature task extends.
+const CONTRACT_TASKS = ['FND-001c', 'FND-002', 'FND-003', 'FND-004']
 const BASE_BRANCH = 'main'
 const MAX_REVIEW_ROUNDS = 2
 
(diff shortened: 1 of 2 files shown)


afternoon:
## Two limitations this PR does not fix

Both surfaced while running the three tasks through the workflow. Recording them here rather than fixing them silently, since each changes workflow policy and has a cost.

### 1. `MAX_REVIEW_ROUNDS = 2` is too low — evidenced 3/3

Every task ran out of review rounds with a real defect still open, and in each case an extra round found something that would otherwise have shipped:

| Task | Rounds used | What the round past the ceiling found |
| --- | --- | --- |
| `FND-001b` | 2, then a manual 3rd | Documented rollback command that fails against the pinned CLI — `hosting:clone` splits the source on `:` before falling back to `@`, so the version ID parses as a channel. A procedure a human follows mid-incident. |
| `FND-007` | 2, then a 3rd | `Tone.Player` divides `duration` by `playbackRate`, so passing song-timeline ticks truncated or overran every loop whose `sourceTempo` differed from the song. Offline and stem export inherit it. |
| `FND-001c` | 2, then a 3rd and 4th | Three findings, including: the opt-out did not stop GA4 automatic collection while the disclosure told users *"Turning this off stops collection."* |

The last two passed a fully green suite — 768 and 954 tests respectively — because each bug lived on a path no fixture exercised (every audio fixture authors at `sourceTempo === tempo`; consent was only ever checked at init on a deep-linked session).

Raising the ceiling costs an Opus review round per task that needs it. On this evidence that looks cheap relative to what the rounds caught, but it is a real cost and a deliberate call.

A cheaper variant worth considering: keep the ceiling at 2 for the *first* pass and allow a 3rd only when round 2's findings are new rather than repeats — the signal in all three cases was that round 2 found a **different** defect, not that it re-raised an unfixed one.

### 2. `runTask` always branches from `main`, so dependent tasks can't be stacked

`BASE_BRANCH` is a module constant baked into `runTask`, so a task can only start once its dependency has **merged**. When `FND-001c` needed `FND-001b`'s deploy pipeline and #25 was still open, the workflow could not express that — branching from `main` would have handed the implementer a tree with no pipeline in it and produced a second, parallel one. It had to be run from a one-off script with the base overridden.

This will recur constantly in Phase 1, where most tasks declare dependencies on tasks that won't have merged yet. A per-task `base` (defaulting to `BASE_BRANCH`), or a base override in `args`, would remove the need for one-off scripts and keep the review loop and PR step intact.

---

PR #111, opened 2026-08-03
LOOP-004 - Synth and one-shot sampler

### LINKED ISSUE #44: LOOP-004 - Synth and one-shot sampler
**LOOP-004 — Synth and one-shot sampler**

| Field | Value |
| --- | --- |
| Phase | 1 — complete the loop workflow |
| Dependencies | `LOOP-003`, `CNT-001` |
| PRD | INS-01 |

Implement reusable synth and sampler graph/UI slices with presets, audition, pitch, sample start/end, envelopes, oscillator, and resonant filter.

## Acceptance criteria

- [x] Controls dispatch validated commands and parameter changes reuse nodes with safe smoothing.
- [x] Sample replacement cancels stale loads and preserves compatible clip data through undo/redo.
- [x] Audio fixtures test parameter extremes, repeated triggers, cache ownership, export compatibility, and disposal.
- [x] Emits `instrument_changed` with `instrument_type` and the `synth` and `sampler` `feature_first_use` keys. Continuous parameter gestures emit nothing per tick.

This issue is the live record for LOOP-004: it is the specification as well as the status. Status is its state and labels, ownership is its assignee, and progress, blockers, and evidence are comments. Product behavior is authoritative in [`docs/prd.md`](../blob/main/docs/prd.md).

afternoon:
## Summary

Implements the INS-01 alpha synth and one-shot sampler as reusable graph/UI slices — parameters, validated commands, real audio voices, and the panel UI, wired into the editor with audition.

- **Domain**: registers the synth (oscillator waveform, amp ADSR, resonant low-pass cutoff/resonance) and sampler (pitch, sample start/end, amp ADSR) parameters once in `src/domain/parameters.ts`, namespaced by instrument kind, and validates stored instrument parameters (plus a sample start<end invariant) in `parse.ts`.
- **Commands**: adds an `instrument` scope to `parameter.set`, plus two new registered commands — `instrument.change` and `instrument.setSample` — with generated inverses, so control edits and sample replacement are undoable and clip data survives instrument swaps.
- **Audio**: rebuilds the `InstrumentGraph` synth voice (PolySynth + resonant `Tone.Filter`) and sampler voice (playback-rate pitch, start/end windowing, per-trigger amp envelope). Parameter edits reuse nodes with click-free ramped smoothing; a sample swap cancels the stale decode via the buffer cache's generation tracking. Adds `ProjectAudioGraph.auditionTrack`, exposed through `useProjectAudio`.
- **UI**: reusable `FillSlider`, `OptionGroup`, `SynthPanel`, and `SamplerPanel` under `src/instrument`, following the design mocks' vertical fill-slider and labelled option-group language, wired into `EditorView` with audition.
- **Analytics**: emits `instrument_changed` (`instrument_type`) on sample replacement and the `synth`/`sampler` `feature_first_use` keys; continuous slider gestures commit once and emit nothing per tick. Pins the two new command ids in the analytics catalog.

## PRD requirements

Implements **INS-01 — Electronic instruments (P0)** ([`docs/prd.md`](../blob/main/docs/prd.md#L341)). The design-mock controls called out as post-alpha in the PRD note (sub-oscillator, pulse-width, multi-mode filter, filter envelope amount, key tracking, sampler fine tune/gain/pan/envelope hold/per-sampler filter) are explicitly out of scope for this alpha baseline and deferred to P1/P2 (tracked in `docs/backlog.md`).

## Acceptance criteria met

- [x] Controls dispatch validated commands and parameter changes reuse nodes with safe smoothing.
- [x] Sample replacement cancels stale loads and preserves compatible clip data through undo/redo.
- [x] Audio fixtures test parameter extremes, repeated triggers, cache ownership, export compatibility, and disposal.
- [x] Emits `instrument_changed` with `instrument_type` and the `synth` and `sampler` `feature_first_use` keys. Continuous parameter gestures emit nothing per tick.

Closes #44

---

### DIFF
--- a/src/analytics/catalog.ts
+++ b/src/analytics/catalog.ts
@@ -182,6 +182,8 @@ export const COMMAND_IDS = [
 	"placement.delete",
 	"placement.update",
 	"parameter.set",
+	"instrument.change",
+	"instrument.setSample",
 ] as const;
 export type CommandId = (typeof COMMAND_IDS)[number];
 
--- a/src/audio/InstrumentGraph.test.ts
+++ b/src/audio/InstrumentGraph.test.ts
@@ -3,8 +3,13 @@ import type { DrumPad, Instrument } from "../domain/entities";
 import { packVersion } from "../domain/entities";
 import type { AssetId, PadId } from "../domain/ids";
 import { createSeededIdFactory } from "../domain/ids";
+import {
+	SAMPLER_PITCH,
+	SYNTH_FILTER_CUTOFF,
+	SYNTH_WAVEFORM,
+} from "../domain/parameters";
 import type { AudioAssetProjection } from "../projection/audioProjection";
-import { installWebAudioGlobals } from "./testAudioContext";
+import { hfEnergy, installWebAudioGlobals, rms } from "./testAudioContext";
 
 installWebAudioGlobals();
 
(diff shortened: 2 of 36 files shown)


afternoon:
CI fix pushed (`4bf6ccc`).

**Root cause:** the unit run failed with a single uncaught `TypeError: Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'`, attributed to `InstrumentGraph.test.ts`. It's a teardown race, not a test-logic bug: `node-web-audio-api` fires a scheduled source's `ended` event from a **native** callback on the host event loop. When a synth/sampler test schedules a note whose stop time is still in the future and then disposes/closes the context, that callback can fire *after* the whole Vitest run has finished and jsdom is torn down — and `propagateEvent` then calls `dispatchEvent(new Event(...))` with a now-stale global `Event`, which the bound jsdom `dispatchEvent` rejects. In isolation the file passes; only the full-suite timing surfaces it, hence the flake.

**Fix:** a narrowly-scoped `uncaughtException` guard (`src/audio/testAudioTeardown.ts`), wired through `installWebAudioGlobals()` so every Tone-touching suite installs it once. It swallows **only** this artifact — matched by a `node-web-audio-api .../events.js` stack frame *plus* the `dispatchEvent`/`Event` type-mismatch message — and re-throws every other uncaught error, so real failures still fail the run. The collision can only happen after teardown, so ignoring it cannot mask an in-test assertion.

**Verified:** reproduced the exact `propagateEvent → dispatchEvent` collision deterministically; confirmed the guard swallows it *and* still crashes on an unrelated uncaught error; full `bun run test` is green (114 files / 1495 tests, 0 errors) where it previously reported 1 error. `typecheck` and `check:ci` pass (the one remaining Biome warning in `sync-issue-dependencies.test.mjs` is pre-existing and unrelated).

afternoon:
My first fix (`4bf6ccc`) didn't hold — CI failed again at the identical spot. Corrected in `3e134b1`.

**Why the first attempt failed:** it added a `process.on("uncaughtException")` listener that ignored the artifact. But Vitest installs its *own* `uncaughtException` listener and records the error into the run's unhandled-errors list; a peer listener can't stop Vitest's from also firing, so the run still failed.

**Correct fix:** stop the throw at its source. The stray event is `node-web-audio-api` firing a scheduled source's `ended` from a native callback after jsdom teardown — `propagateEvent` then calls `dispatchEvent(new Event(...))` with a now-stale global `Event`, which the bound jsdom `dispatchEvent` rejects. I wrap `node-web-audio-api`'s own `AudioNode.prototype.dispatchEvent` (an *own* property — the global `EventTarget` is untouched) to swallow **only** that post-teardown stale-`Event` TypeError and rethrow everything else. The event never throws, so nothing reaches any `uncaughtException` listener. Predicate matches structurally on name/message because `instanceof Error` is unreliable across the jsdom realm boundary.

**Verified against the real flake, not a mock:**
- New `src/audio/testAudioTeardown.test.ts` drives the actual jsdom + node-web-audio-api dispatch path and asserts both the swallow and that a well-formed event still dispatches.
- Instrumenting the guard with a counter showed it swallowing **51 real stale-`Event` dispatches** across one full `bun run test`, which now reports **0 errors** (115 files / 1497 tests).
- `typecheck` and `check:ci` pass (the lone Biome warning in `sync-issue-dependencies.test.mjs` is pre-existing and unrelated).

PR #135, opened 2026-08-04
HARD-006 - Detailed performance audit and fixes

### LINKED ISSUE #127: HARD-006 Detailed performance audit and fixes

afternoon:
## What & why

`HARD-006` is a detailed performance audit of the per-edit and per-frame hot
paths against [PRD section 10 performance budgets](docs/prd.md#performance-budgets)
and [section 9.3 arrangement acceptance](docs/prd.md#93-arrangement-renderer-decision),
plus the fix the audit surfaced. Full method and findings are recorded in
[`docs/perf/hard-006-performance-audit.md`](docs/perf/hard-006-performance-audit.md).

The audit walked the two paths that run on every project change and are
designed to be *incremental* (structural sharing means an unchanged entity
should arrive as the same object reference):

1. **The audio projection** (`src/projection/audioProjection.ts`), rebuilt by
   `useProjectAudio`'s effect on every open-project change — on the "pointer-driven
   note and parameter edits target a visual response within 50 ms" path.
2. **The arrangement projection** (`src/projection/arrangementProjection.ts`),
   rebuilt whenever the renderer's project revision moves — feeding the section
   9.3 acceptance that frame cost is proportional to visible objects, not total
   project size.

**Finding:** both builders correctly reused unchanged entities, but decided
reuse by computing a content fingerprint (stable-stringify + hash)
*unconditionally* for every entity before comparing digests. On the 50-track,
2,500-placement reference project, a single one-note edit re-serialized and
re-hashed thousands of untouched tracks, placements, clips, automation lanes,
returns, and assets — cost proportional to total project size, exactly what
section 9.3 forbids, and squarely on the 50 ms pointer-edit budget since it
runs synchronously inside the edit's effect.

**Fix:** `src/projection/fingerprint.ts` gains `rememberSource`/
`reuseIfUnchanged`, an object-identity fast path backed by a `WeakMap` keyed
on the previously built projection entry. Each builder now checks whether its
source inputs are `Object.is`-identical to what the previous entry was built
from, and if so returns the previous entry immediately, skipping the
fingerprint entirely; it only falls through to the existing fingerprint path
when a dependency actually differs, so correctness never rests on the fast
path. The `WeakMap` adds no field to the published `AudioSongProjection`/
`ArrangementProjection` shapes, so this is not a contract change — only a
speed-up.

The audit also confirmed several budgets already in spec with no code change
needed (arrangement range/row binary searches, the audio graph's per-entity
reconciliation, the waveform LRU cache, the monitoring SDK being off the
interactive path) — see the audit doc for detail.

No new user-facing action is introduced, so no new OPS-02 event; the audio
hot path this touches is already covered by the existing `audio_underrun`
reliability event.

Closes #127

## Walkthrough

No UI change. This is an internal fix to the projection builders' reuse
decision path (`src/projection/audioProjection.ts`,
`src/projection/arrangementProjection.ts`, `src/projection/fingerprint.ts`);
nothing user-visible changed.

## Evidence

```
$ bun run typecheck
$ tsc --noEmit
(clean, no output)

$ bun run test
 Test Files  119 passed (119)
      Tests  1570 passed (1570)
   Duration  88.51s

$ bun run check:ci
$ tsc --noEmit && biome check
Checked 351 files in 489ms. No fixes applied.
```

`src/projection/incrementalReuse.test.ts` counts `fingerprintOf` invocations
across an incremental rebuild of the reference project and asserts the count
scales with *changed* entities, not project size: zero for an unchanged
rebuild, two for a one-track mixer edit (full + topology), one for a
one-placement move. Confirmed to fail on the pre-fix behavior (disabling the
fast path drives the counts back into the thousands) and pass after.
`src/projection/fingerprint.test.ts` covers the new helpers directly,
including keying on identity rather than deep equality and degrading
gracefully for an entry that predates the cache. The existing
`audioProjection.test.ts`/`arrangementProjection.test.ts` reuse suites pass
unchanged, confirming reused-entry identity and genuinely-changed-entry
correctness are untouched.

## Acceptance criteria met

- [x] Audited the per-edit hot path (audio projection rebuild on every
      project change) against the PRD section 10 50 ms pointer-edit budget.
- [x] Audited the per-frame-adjacent hot path (arrangement projection
      rebuild) against the PRD section 9.3 "frame cost proportional to
      visible objects, not total project size" acceptance.
- [x] Fixed the finding: projection builders no longer re-fingerprint
      unchanged entities; reuse decisions take an object-identity fast path
      ahead of the content fingerprint.
- [x] Fix is guarded by a regression test that fails on the pre-fix
      behavior and passes after (`incrementalReuse.test.ts`).
- [x] No contract change — `AudioSongProjection`/`ArrangementProjection`
      published shapes are unchanged; existing projection contract tests
      pass unchanged.
- [x] Audit findings and method recorded in `docs/perf/hard-006-performance-audit.md`.
- [x] `bun run typecheck`, `bun run test`, `bun run check:ci` all pass.
- [ ] Not claimed: the section 9.3 frame budgets measured on the physical
      baseline device in the gating browsers. That binds at
      `ARR-005`/`HARD-001` on that device; this task does not gate on
      hardware or a hosted environment that isn't provisioned yet.

---


afternoon:
@claude Fix tests

PR #114, opened 2026-08-03
LOOP-013 - Searchable, sync-audition library browser

afternoon:
## What & why

Implements the `src/library` searchable, sync-audition library browser: a ~250px left-hand **tree panel** you keep open while you work, a **pack browser modal** as the way to browse and add packs, cross-pack search and facet filtering, lazy manifest loading, sync-aware audition through the shared audio runtime, isolated error handling for missing/corrupt assets and packs, and the analytics wired through it.

- `src/library/manifest.ts` — parsed read model of the pack index and per-pack manifests; resolves served-root asset URLs, maps the wire asset types onto the browser's types, and tolerates a preset's `audio: null` so one bad preset never fails a whole manifest.
- `src/library/libraryClient.ts` — lazy fetching: the compact pack index once, a pack's full manifest only when that pack is opened, cached by immutable pack+version. A failed pack is an isolated, retryable error and never takes down the packs that loaded fine.
- `src/library/tree.ts` — the panel's pack → role-group → asset tree. The manifests already carry the generator's `family`/`role` taxonomy, so "Kicks / Claps / Closed hats" is derivable today without new pack metadata.
- `src/library/search.ts` — pure search plus genre/role/character/type facets (AND across facets, OR within a facet); clearing the pack facet only widens results, and clearing a genre filter re-exposes everything it had hidden.
- `src/library/audition.ts` + `src/library/toneAuditionEngine.ts` — sync-aware audition (loops sync to the transport, one-shots fire immediately) through the shared `AudioRuntime` destination, never an export/offline context. One preview plays at a time and stops on selection change, close, navigation, or project teardown; the policy layer is Tone-free and unit-tested, with `toneAuditionEngine.ts` as the one Web-Audio-touching file.
- `src/library/useLibraryBrowser.ts` + `LibraryBrowser.tsx` + `PackBrowser.tsx` — the reactive controller, the panel, and the modal, with keyboard navigation, loading/error states, and an `onInsert` affordance; wired into `EditorView` behind a header toggle.
- `src/analytics/catalog.ts` — extends `library_audition` with a stable `pack_id` (the pack slug, never its display name, pinned against the delivered pack set by `catalog.test.ts`), adds `library_pack_added`, plus `asset_load_failed` and the `library_browser` / `pack_browser` `feature_first_use` keys.

PRD requirements: **LIB-01, LIB-02, LIB-04**.

Closes #54

### Scope decisions carried from the issue

[...]

PR #137, opened 2026-08-04
LIB-08 - Persist the packs a project has added, behind a command

### LINKED ISSUE #131: LIB-08 - Persist the packs a project has added, behind a command
## Why

`LOOP-013` (#114) gives the library panel a tree whose top level is "the packs this project has", and the pack browser a primary **Add pack to project** action. There is no project state behind that today.

`metadata.packDependencies` is *derived* from the assets a project actually uses (`derivePackDependencies`, recomputed once per transaction and validated by `parseProject`). It answers "which packs does this project need in order to open?" — not "which packs has the user put on their shelf?". A user can legitimately add a pack, browse it, and not yet use a sound from it; deriving the shelf from usage would make that pack vanish on reload.

So the browser currently seeds the tree from `packDependencies` and keeps anything added afterwards as **editor-session state** (`sessionPackIds` in `EditorView.tsx`). It survives the session and nothing more. That is honest but temporary: the point of adding a pack is that it is still there tomorrow.

## Scope

This is a **domain contract change** and therefore its own task, not incidental feature work (see `CLAUDE.md`, "A contract-owning task lands before its dependents start"):

- Decide where the shelf lives: a project-level list of added packs alongside the derived dependency list, and whether it is per-project or per-user (a per-user library is `LIB-03` territory; per-project is the smaller, alpha-shaped answer and matches how a project already records its dependencies).
- Add the field to schema v1's project metadata, its Zod schema, and its invariants — including how it relates to `packDependencies` (a used pack must appear in both; an added-but-unused pack appears only in the shelf).
- Register the command(s) that mutate it (`pack.add` / `pack.remove`), with inverses, summaries, and their entries in the pinned command registry test and the analytics `COMMAND_IDS` list.
- Persist it in the metadata tier, and add the forward migration for projects saved before it.
- Replace `EditorView`'s `sessionPackIds` with a dispatch through the command layer, and make `library_pack_added` fire from that path.

## Acceptance criteria

- [ ] A pack added in the pack browser is still in the project's panel after a reload, and after the project is reopened on another device.
- [x] Adding and removing a pack goes through a registered command, is undoable, and produces exactly one revision per action.
- [x] An added pack that the project uses no asset from is preserved; removing a pack whose assets are in use is either refused with a named reason or reported as a missing-pack state (`resolvePackAvailability`) — not silently dropped.
- [x] `parseProject` rejects a shelf that contradicts the derived `packDependencies`.
- [x] Projects saved before this field migrate forward without a user-visible failure.
- [x] `docs/persistence.md` records the field, and the command registry test pins the new commands.

Follows #54.

afternoon:
## What & why

`metadata.packDependencies` is *derived* from the assets a project actually uses — it answers "which packs does this project need in order to open?", not "which packs has the user put on their shelf?". Today, adding a pack in the library panel that hasn't been used yet lives only in editor-session state (`sessionPackIds`) and disappears on reload. This PR gives that shelf a real, persisted home behind a command, so `LOOP-013`'s pack browser has project state to write to.

Satisfies PRD `LIB-05` (pack dependency derivation and missing-pack state) by adding the complementary maintained-not-derived shelf alongside it, per the task's scope in issue #131.

**What shipped:**
- `metadata.addedPacks: PackDependency[]` (schema **v2**) — a project-level, maintained (not derived) list of packs the user has added. `parseProject` enforces that every pack backing a used asset (`packDependencies`) also appears in `addedPacks` at the same version; a used pack contradicting the shelf, or a version mismatch, is rejected. `withDerivedPackDependencies` reconciles the shelf automatically when an edit introduces an asset from a new pack.
- `pack.add` / `pack.remove` commands — registered, undoable, one revision per action, with inverses and summaries pinned in the registry test and `COMMAND_IDS`. `pack.remove` is refused with a named reason when the pack still backs an asset in use, rather than silently dropping it.
- Persistence: `ProjectMetadataPatch` carries `addedPacks`; `saveSong` reconciles it against the derived list in the same revision. A `v1 -> v2` forward migration seeds the shelf from a project's existing dependencies so pre-existing projects open and list unchanged. `decodeStoredProject` / `decodeStoredProjectMetadata` migrate at load and dashboard-list time.
- `firestore.rules` allowlists `addedPacks` and moves to `schemaVersion == 2`; covered by the rules emulator test.
- Analytics: `library_pack_added` (`pack_kind`) added to the OPS-02 catalog and the PRD table, with once-per-action and consent tests, ready for `LOOP-013` to call.

**Deliberate scope boundary:** the pack-browser UI and `EditorView`'s `sessionPackIds` do not exist in the tree yet (`LOOP-013` / #114, which owns the browser and the `Pack` record, has not landed). This PR lands the full domain/command/persistence/migration/analytics contract; wiring the browser's *Add pack* action to `pack.add` and firing `library_pack_added` from it is `LOOP-013`'s job, noted on the issue as a hand-off.

Closes #131

## Walkthrough

No UI change. This is a domain/command/persistence contract change with no reachable UI surface yet — the pack browser that will call `pack.add`/`pack.remove` ships in `LOOP-013` (#114), which hasn't landed. No component, route, or editor UI file is touched by this PR.

## Evidence

- `bun run typecheck` — pass
- `bun run check` (and `check:ci`) — pass
- `bun run test` — pass (1582 tests)
- Firestore rules emulator suite: updated in lockstep with `firestore.rules` and exercised via the shared `ProjectRepository` contract suite against the in-memory store (identical encode/decode/contract logic to the Firestore store); the emulator run itself could not execute in the implementation environment (no Java runtime available for `firebase emulators:exec`) and should be confirmed in CI.
- This branch passed an Opus review round in the implementation workflow.

## Acceptance criteria met

- [x] Adding and removing a pack goes through a registered command, is undoable, and produces exactly one revision per action.
- [x] An added pack that the project uses no asset from is preserved; removing a pack whose assets are in use is refused with a named reason rather than silently dropped.
- [x] `parseProject` rejects a shelf that contradicts the derived `packDependencies`.
- [x] Projects saved before this field migrate forward without a user-visible failure (`v1 -> v2` migration, exercised by `migrations.test.ts`).
- [x] `docs/persistence.md` records the field, and the command registry test pins the new commands.
- [ ] "A pack added in the pack browser is still in the project's panel after a reload, and after the project is reopened on another device." — the persistence guarantee underlying this is proven by a new `ProjectRepository` contract test run by both the in-memory and Firestore stores. The box names the pack browser and project panel, which are `LOOP-013` (#114); that task will wire the UI onto the command landed here and can then close this box.

---


afternoon:
@claude I'd like to create logs that allow me to see which packs are the most popular. This can be privacy-sensitive, we don't need to tie that to user ID or PII. If it is not currently possible for me to query analytics data for this, add it.

afternoon:
@claude I would like to be able to see pack keys for third-party packs as well. Assumption for user packs is that they are not shareable (otherwise they become third-party right?), so its correct not to log pack keys for those. The reason for logging third party pack keys is that we want to provide analytics on which packs are successful, for internal use and to give feedback and validation to third party pack creators. Add support for this. And fix merge conflicts!

afternoon:
@claude More code has merged in the meantime. Please fix merge conflicts again.

Issue #48, opened 2026-07-27
LOOP-008 - Device-chain and routing framework


afternoon:
**LOOP-008 — Device-chain and routing framework**

| Field | Value |
| --- | --- |
| Phase | 1 — complete the loop workflow |
| Dependencies | `LOOP-007` |
| PRD | FX-01, FX-02 |

Implement typed ordered devices, eight inserts per track, master devices, two stereo returns, sends, bypass/reset/duplicate/reorder, preset provenance, and parameter widgets.

## Acceptance criteria

- [ ] Device/routing commands preserve invariants, history, autosave, and stable IDs.
- [x] Reorder reuses nodes with a measured click-safe reconnection strategy; unrelated graphs retain identity.
- [x] Extreme but finite ranges remain creative while invalid values and unsafe feedback are rejected.
- [ ] Emits `device_added` with `device_type` and `chain` (insert, return, or master), plus the `device_chain` and `send_return` `feature_first_use` keys.

This issue is the live record for LOOP-008: it is the specification as well as the status. Status is its state and labels, ownership is its assignee, and progress, blockers, and evidence are comments. Product behavior is authoritative in [`docs/prd.md`](../blob/main/docs/prd.md).

afternoon:
Starting LOOP-008 (device-chain and routing framework).

Branch: `claude/loop-008` off `origin/main`.

Plan: typed device commands (add/remove/reorder/duplicate/bypass/reset) plus send create/remove through the shared command layer with generated inverses and stable IDs; concrete device parameter definitions and audio node factories for the six core device types (filter/EQ, overdrive, saturation, compression, delay, reverb) wired through the existing `DeviceChain` reconciliation; domain invariants for the eight-insert-per-track ceiling and the two-return limit, with extreme-but-finite parameter ranges and rejection of non-finite/unsafe values; `device_added` (with authored `device_type` keys + `chain`) plus `device_chain`/`send_return` `feature_first_use`; and a device-chain / send UI panel in the editor.

---
_Generated by [Claude Code](https://claude.ai/code)_

afternoon:
Pushed the domain, command, analytics-catalog, and click-safe-audio slices to `claude/loop-008` (5 commits on `origin/main`). Full unit suite green: 1643 tests, `bun run typecheck` and `bun run check` clean.

**Delivered**

[...]

PR #138, opened 2026-08-04
LOOP-008 (1/4): device-type registry and capacity invariants

afternoon:
## What & why

**PR 1 of 4 in the LOOP-008 stack** (device-chain and routing framework). This is the domain-contract slice, landed **contract-first** per CLAUDE.md rule 4 so its command dependents (PR 2) build on a published contract rather than altering one as incidental work.

Introduces the typed device-type registry (`src/domain/devices.ts`) — the six core device types with extreme-but-finite parameter ranges — and the capacity invariants in `src/domain/parse.ts` (eight inserts per track, two stereo returns), with their tests.

Satisfies PRD **FX-01**, **FX-02** (the domain half).

Part of #48 — this PR does not close the issue on its own; the stack does.

### Why this is one PR despite exceeding 400 lines
This is a single domain contract (registry + the invariants that guard it). The registry table and its parameter definitions are a coherent unit that consumers read as one thing; splitting the type table from the capacity invariants that reference it would land a half-contract. It is landed first and alone so a reviewer holds only the contract in their head.

## Walkthrough

No UI change. Domain model, invariants, and tests only.

## Evidence

Run on this branch:
- `bun run typecheck` — passes.
- `bun run check:ci` — `Checked 384 files. No fixes applied.`
- `bun run test` — 132 files, 1714 tests passed.

## Acceptance criteria met

- [x] Extreme but finite ranges remain creative while invalid values and unsafe feedback are rejected. (device parameter ranges + capacity invariants)

The remaining LOOP-008 acceptance boxes are satisfied by the later PRs in the stack.

### DIFF
--- a/src/domain/devices.test.ts
+++ b/src/domain/devices.test.ts
@@ -1,0 +1,115 @@
+import { describe, expect, it } from "vitest";
+import {
+	createDevice,
+	defaultDeviceParameters,
+	deviceParameters,
+	deviceTypeDefinition,
+	deviceTypes,
+	isRegisteredDeviceType,
+} from "./devices";
+import { createSeededIdFactory } from "./ids";
+import {
+	bareParameterId,
+	coerceParameterValue,
+	getParameterDefinition,
+	isParameterValueInRange,
+} from "./parameters";
+
+const deviceId = createSeededIdFactory("devices-test")("device");
+
+describe("device type registry", () => {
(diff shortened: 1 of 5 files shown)


afternoon:
@claude The limits on device chain length and return count are too restrictive. Lets increase device chain max length to 16 and return count to 8.

PR #141, opened 2026-08-04
LOOP-008 (4/4): click-safe device-chain relink

afternoon:
## What & why

**PR 4 of 4 in the LOOP-008 stack — independent, based directly on `main`.** Has no code or file overlap with PRs 1–3 and can be reviewed and merged in any order relative to them.

Makes `DeviceChain`'s reorder relink click-safe: reordering an insert chain reuses existing nodes and reconnects them without recreating anything, so unrelated device graphs retain node identity and playback is not interrupted (`src/audio/DeviceChain.ts` + its test). Operates only on the passthrough nodes Schema v1 uses today; no concrete DSP.

Satisfies PRD **FX-01** (the audio-layer reconnection guarantee).

Part of #48.

**Why its own PR:** this is the only audio-layer change in LOOP-008. It shares no file and no dependency with the domain contract, command family, or analytics claim, so folding it into any of them would force those reviewers to re-review an unrelated audio concern.

## Walkthrough

No UI change. Audio-graph reconnection strategy and its test only.

## Evidence

Run on this branch (based on `main`):
- `bun run typecheck` — passes.
- `bun run check:ci` — `Checked 382 files. No fixes applied.`
- `src/audio/DeviceChain.test.ts` — 9 tests passed.
- Full `bun run test` — 131 files, 1705 tests passed.

## Acceptance criteria met

- [x] Reorder reuses nodes with a measured click-safe reconnection strategy; unrelated graphs retain identity.

### DIFF
--- a/src/audio/DeviceChain.test.ts
+++ b/src/audio/DeviceChain.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, beforeAll, describe, expect, it } from "vitest";
 import type { Device } from "../domain/entities";
+import { createManualScheduler } from "../shared/scheduler";
 import { installWebAudioGlobals, rms } from "./testAudioContext";
 
 // Must run before Tone is imported — see AudioRuntime.test.ts for why.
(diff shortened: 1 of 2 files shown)


afternoon:
**Re-review after fix: APPROVED.**

The prior round blocked this slice because `DeviceChain.relink()` performed the `disconnect()`/`connect()` topology change synchronously while output gain was still ~1 — the fade to zero completed 5ms *after* the rewire, so it masked nothing ("make-before-break that breaks before it makes"). Fixed in `33203b0`: the rewire is now deferred via an injectable `Scheduler` until after the down-ramp reaches silence, then gain ramps back to 1.

An Opus reviewer adversarially verified the fix and confirmed:
- The disconnect/connect genuinely cannot run while gain > 0 (deferred by `RELINK_FADE_SECONDS`; timers fire late, never early).
- No new hazard introduced — gain-stuck-at-0, double-connected/orphaned wiring on a mid-fade reorder, a pending rewire firing after `dispose()`, and leaked scheduled tasks are each handled (coalescing cancels the superseded relink; `rewire()` reads live order and fully rebuilds; `dispose()` cancels any pending task).
- The new ordering test asserts the rewire happens *during silence* (event ordering), not merely that ramp values are `[0,1]` — and it fails 5/13 against the pre-fix code.
- Constructor change is additive (`scheduler` defaults to `timeoutScheduler`); all three call sites unchanged.
- Gates: typecheck ✓, `bun run test` 1709 passed (DeviceChain 13/13) ✓, `check:ci` ✓.

---

PR #164, opened 2026-08-05
Add CONTRIBUTING.md as the local development guide

afternoon:
## What

Adds `CONTRIBUTING.md` as the entry point for local development, and documents the one workflow that had no written instructions anywhere: **driving the app by hand against the Firebase Emulator**.

`docs/testing.md` covered the two automated emulator suites (`test:emulator`, `test:browser:emulator`), both of which start and stop their own emulator. Neither helps when you want to click around the real app against real persistence and real security rules. The README offered only two modes — a real Firebase project, or `VITE_MOCK_BACKEND=true` — and the mock is structurally unable to show anything about persistence, since every page load gets a fresh empty store.

## The recipe is verified, not inferred

I ran it end to end on this branch rather than deriving it from `playwright.emulator.config.ts`:

1. `firebase emulators:start --only firestore,auth --project demo-solid-groove`
2. `bun run dev` with the five documented env vars
3. Landing page → "Start in your browser" → anonymous user appears in the Auth emulator (`accounts:query` reports `recordsCount: 1`)
4. New Project → editor loads the starter project (BD track, 4 steps)
5. Toggled a fifth step → clip document in Firestore emulator holds the notes
6. **Reloaded the page → the fifth step came back from Firestore**, then restarted the dev server entirely and reloaded again to confirm the minimal env set is sufficient

That last step is the one the mock backend cannot prove.

Traps the exercise surfaced, all now documented:

- `VITE_MOCK_BACKEND` must not be `"true"` — the code compares against that exact string.
- The three `VITE_FIREBASE_*` values **are** required and their contents don't matter. `src/firebaseConfig.ts` only substitutes placeholders in mock mode, so outside it `initializeApp` gets `undefined`. My first answer to the user omitted this.
- The Emulator UI is disabled in `firebase.json`, so browsing data means the REST API with the owner bearer token.
- A plain `curl` against that API returns `PERMISSION_DENIED` — that is `firestore.rules` working, not a failed write. Easy to misread as a bug.
- `127.0.0.1`, not `localhost`: both emulators listen on the IPv4 loopback only (confirmed with `lsof`).

## Duplication removed

- **The null-ALSA setup block** was verbatim in `CLAUDE.md`, while `docs/testing.md` explained the reasoning and deferred the actual config to "see the CI step for the exact config" — so the copy-pasteable steps existed in exactly one place, and it wasn't the testing doc. The steps now live in `CONTRIBUTING.md`; both files link to them. `CLAUDE.md` keeps the two warnings that are load-bearing for an agent reading it: don't "fix" the error by mocking Tone, and it does not apply to the browser suites.
- **README getting-started** no longer restates the mock-backend explanation.
- **`.env.example`**'s emulator block pointed at a partial command; it now points at the full recipe and notes the `VITE_FIREBASE_*` requirement.

`docs/testing.md` remains the suite-and-CI reference and now says so explicitly, so the two documents frame each other instead of overlapping.

## UI walkthrough

No user-visible change — documentation and `.env.example` comments only.

## Evidence

- `bun run check`: `biome` clean. `tsc --noEmit` reports one error in `src/arrangement/useArrangementCanvas.test.ts` (`createArrangementSpikeProject` no longer exported from `domain/fixtures`) — **pre-existing and unrelated**, reproduced on an unmodified `main` checkout. It's a stale test left from the removed arrangement spike; not fixed here to keep this diff to its one purpose.
- `src/shortcuts/docs.test.ts`: 6/6 pass (the docs-drift gate).
- All internal markdown links and anchors across the four touched files resolve, checked with a script.
- Diff is 5 files, well under the 400-line ceiling.

Worth a follow-up issue for that stale arrangement test if one isn't already open.


afternoon:
Superseded by #171 (docs) and #170 (the `VITE_DEV_BACKEND` contract). This PR shared a branch with the follow-up work, so it could not show its own diff; splitting it gives each half a reviewable one.

Issue #166, opened 2026-08-05
HARD-006 - Session Replay disclosure, purpose, and opt-out


afternoon:
**HARD-006 — Session Replay disclosure, purpose, and opt-out**

| Field | Value |
| --- | --- |
| Phase | Alpha Milestone 4: Private Alpha Hardening |
| Decided by | [ADR 0002](../blob/main/docs/adr/0002-sentry-session-replay.md) |
| Needed by | `HARD-003`, `HARD-005`, `OPS-001` |

[ADR 0002](../blob/main/docs/adr/0002-sentry-session-replay.md) adopts Sentry Session Replay to understand how people use the app, superseding [ADR 0001](../blob/main/docs/adr/0001-sentry-for-error-monitoring.md) decision 4. The decision is recorded; the app does not yet implement it. This task makes the app honest about it and gives the user control.

**Replay must not ship before this task lands.** Enabling capture without the disclosure and the opt-out would collect session recordings the user was never told about — exactly the failure the PRD section 15 risk row names. If replay capture is enabled in the same stack, it is the *last* PR in it, after disclosure and opt-out are in place.

## Scope

**Disclose it, and say what it is for.** `src/components/TelemetryDisclosure.tsx` currently names two processors and promises "no notes or audio, and no text you type". Replay makes that copy incomplete. The disclosure must name Session Replay specifically and say, in plain language:

- that it records how the app is used — which controls are clicked, how the user moves through the app;
- that its purpose is to understand how people use Solid Groove so it can be improved;
- that it is **not** used to access the user's music or other private information;
- that project content is masked out at capture, so it never reaches Sentry at all.

The existing promise that the user's music never leaves their project must remain literally true with replay on. If the masking cannot support that sentence, the masking is wrong — not the sentence.

**Extend the opt-out to cover replay.** `src/analytics/consent.ts` models two processors (`productAnalytics`, `errorMonitoring`). Add a third flag for session replay so the preference is expressible on its own, while the single user-facing control keeps turning everything off in one action, as it does today. Declining must stop replay **at the source** — the integration does not capture — rather than dropping a recording before send. Opting out mid-session stops an in-flight recording. `CONSENT_DEFAULT` gains the new flag; its default value belongs to `DEC-009` (#38) alongside the others, and is set the same way — one line here.

[...]

Issue #38, opened 2026-07-27
DEC-009 - Analytics consent and retention policy


afternoon:
**DEC-009 — Analytics consent and retention policy**

| Field | Value |
| --- | --- |
| Phase | Product decision |
| Needed by | `LOOP-001b`, `HARD-003` |

Decide whether product analytics are on by default with an opt-out or require opt-in, what the user-facing disclosure says, how long event and error data are retained, and any regional constraints. The disclosure covers two processors — Google Analytics for product events and Sentry for error monitoring ([ADR 0001](../blob/main/docs/adr/0001-sentry-for-error-monitoring.md)) — and a regional constraint Sentry's SaaS regions cannot meet would reopen the self-hosting option that ADR rejected. `FND-001c` builds the opt-out mechanism and the disclosure hook regardless; this decision sets the default state and the wording. It does not block `FND-001c`, and it must be settled before the cohort is invited in `HARD-005`.

This issue is the live record for DEC-009: it is the specification as well as the status. Status is its state and labels, ownership is its assignee, and progress, blockers, and evidence are comments. Product behavior is authoritative in [`docs/prd.md`](../blob/main/docs/prd.md).

afternoon:
Partial decision recorded — **default state only**.

**Decided:** Session Replay (and the analytics it accompanies) is **on by default with an opt-out**, not opt-in. Approved by @bpgodfrey on 2026-08-06.

This unblocks the `CONSENT_DEFAULT.sessionReplay = true` line shipping in the HARD-006 stack (#189-#194, #199), which was written as a placeholder explicitly pending this decision.

**Still open on DEC-009** — this comment does not settle them:
1. The user-facing disclosure wording for the two processors (Google Analytics for product events, Sentry for error monitoring + replay, per ADR 0001 / ADR 0002).
2. Event and error/replay data retention periods.
3. Any regional constraints. Per the issue body, a constraint Sentry SaaS regions cannot meet would reopen the self-hosting option ADR 0001 rejected.

Leaving this issue open and `human-input-required` until 2-4 are settled. Per the body it must be closed before the cohort is invited in HARD-005.

PR #189, opened 2026-08-06
HARD-006 - Session Replay consent flag (1 of 6)

afternoon:
## What & why

1 of 6 in the HARD-006 stack, builds on `main`.

This is the registration-only slice (per CLAUDE.md "Landing work" item 3): `src/analytics/consent.ts` is a shared registration point, so it lands first and alone, ahead of the masking, SDK configuration, and disclosure slices that build on it.

Adds a third `sessionReplay` flag to the consent model alongside the existing `productAnalytics` and `errorMonitoring` flags:

- `ConsentState` carries `sessionReplay`.
- `CONSENT_DEFAULT` gains the new flag (its final default value is `DEC-009` (#38), same as the other two — this is one line, matching how `productAnalytics`/`errorMonitoring` were seeded).
- `CONSENT_ALL_OFF` / the single user-facing "turn everything off" control now clears `sessionReplay` too, in the same action.
- `parse` validates and survives hostile or absent storage for the new flag, same as the existing two.
- A `replayAllowed` accessor is added alongside the existing processor accessors.

Refs #166. This slice does not close the issue — later slices (masking, SDK configuration, consent-at-source wiring, disclosure copy) are still to come.

## Walkthrough

No UI change. This slice only extends the consent data model (`src/analytics/consent.ts`); nothing renders differently.

## Evidence

- `bun run typecheck` — pass
- `bun run check` — pass

This branch passed an Opus review round in the implementation workflow (a prior review round found ordering and payload-modelling issues elsewhere in the stack; those were fixed before this stack was finalized — see issue #166 comments).

## Acceptance criteria met

From #166:
- [x] `ConsentState` carries a session-replay flag; the single user-facing control still turns everything off in one action; the flag persists and survives hostile or absent storage (`src/analytics/consent.test.ts` extended).

The remaining acceptance criteria (disclosure copy, capture stopped at the source, `sentrySink.ts` configuration, OPS-03 payload rule, masking markers, budget re-measurement) are satisfied by the later slices in this stack.

---

### DIFF
--- a/src/analytics/consent.test.ts
+++ b/src/analytics/consent.test.ts
@@ -1,6 +1,11 @@
 import { describe, expect, it, vi } from "vitest";
 import { hostileStorage, memoryStorage } from "../testing/storage";
-import { CONSENT_DEFAULT, CONSENT_STORAGE_KEY, ConsentStore } from "./consent";
+import {
+	CONSENT_DEFAULT,
+	CONSENT_FLAGS,
+	CONSENT_STORAGE_KEY,
+	ConsentStore,
+} from "./consent";
 
 describe("default state (DEC-009 owns the value, this owns the mechanism)", () => {
 	it("uses the declared default when nothing is stored", () => {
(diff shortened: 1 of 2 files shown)


afternoon:
Stack base changed: this now sits on top of #210 (ADR 0003), which must merge first — GitHub declines a base change while the PR is part of a stack, so the ordering is recorded here rather than in the base field. #210 → #189 → #190 → #191 → #192 → #194 → #199. All six slices have been rebased onto #210 and force-pushed.

PR #190, opened 2026-08-06
HARD-006 - Session Replay component-level masking (2 of 6)

afternoon:
## What & why

2 of 6 in the HARD-006 stack, builds on #189 (`claude/hard-006`).

Adds the ADR 0002 decision 2 component-level block/mask marking: a `src/monitoring/replayPrivacy.ts` vocabulary for Sentry's block/mask class or attribute, applied to every surface in `src/` that renders user-authored or project-derived content (track/clip/section names, the arrangement, the piano roll, the assistant transcript, text inputs, library facets/pack lists, and more), plus `replayPrivacy.test.ts` — a table-driven test that fails when a known content surface loses its marking.

This means a new content surface is masked by both the SDK-level default (added in the next slice) *and* its own markup, per the issue's "belt and braces" requirement.

Refs #166.

## Walkthrough

No UI change. This slice adds Sentry block/mask class names and attributes to existing DOM elements only — no layout, styling, or interaction changes.

## Evidence

- `bun run typecheck` — pass
- `bun run check` — pass

## Acceptance criteria met

From #166:
- [x] Every surface rendering user-authored or project-derived content carries the block/mask marking; a test fails if a known content surface is missing it (`src/monitoring/replayPrivacy.test.ts`).

The remaining acceptance criteria are satisfied by the later slices in this stack.

---


afternoon:
@claude From reading the code, it seems that almost the entire UI is masked or blocked, rendering session replay low value. Remove masking on large blocks, like the arrangement or the mixer. Instead add masking conservatively to elements that are known to have sensitive text content, e.g. project name, track names. If in doubt, reply with a list of masking decisions that you would like me to make and I will confirm or decline each.

afternoon:
@claude  Mask aria-label too.

afternoon:
## Your three open questions are settled — pushed as `aac1ac9`

The narrowing you asked for stands as-is; this adds the two follow-ups you were owed answers on, plus one correction the ADR forces.

**1. `ConfirmDialog` — masked the title only**, which was your own recommendation. The title interpolates a name (`Delete "…"?`); the message is fixed explanatory copy, and a delete confirmation is exactly the friction replay exists to show. Verified the new assertion fails without the mark before it passed.

**2. Asset, pad, and pack names — confirmed unmasked.** These are library facts we ship, identical for every user, and which sound someone reached for is signal worth having. ADR 0003 records that this stops being true the day users can upload or rename their own samples, so it gets revisited rather than silently inherited.

**3. `aria-label`** — landed here, wired to the integration in #191, which is where the `maskAttributes` option lives.

**The correction:** your unblocking of `arrangement-canvas-stack` was justified by canvas capture being off. [ADR 0003](https://github.com/afternoon/solid-groove/pull/210) turns it **on**, so that reasoning is now backwards even though the outcome is unchanged — the stack stays unblocked because recording it is the point. The comment and the test's surface table now say that, including that the section names the renderer draws into the canvas *are* recorded and that masking cannot reach inside a canvas. Worth being exact about, since the old comment implied a protection that no longer exists.

Also: your "`maskAllText: true` is what actually greys the UI, not this slice" was the correct diagnosis. The fix turned out to be canvas capture rather than the unmask-chrome allowlist — the arrangement and piano roll being blank was the real cost, not the greyed labels.

**Evidence** — `typecheck` clean, `check` clean, 39 tests pass in `replayPrivacy.test.ts` + `ConfirmDialog.test.tsx`, 2161/2161 at the stack tip.

PR #191, opened 2026-08-06
HARD-006 - Configure Session Replay per ADR 0002 (3 of 6)

afternoon:
## What & why

3 of 6 in the HARD-006 stack, builds on #190 (`claude/hard-006-masking`).

Configures Session Replay in `src/monitoring/sentrySink.ts` — the only module allowed to import `@sentry/*` — per ADR 0002 decisions 2, 6, and 8:

- `maskAllText`, `blockAllMedia`, canvas capture off (asserted by the absence of `replayCanvasIntegration`, not a flag — the SDK only records canvas when that second integration is present).
- Low `replaysSessionSampleRate`, `replaysOnErrorSampleRate` at 0.
- Loaded lazily after first paint, never on the landing page.
- A `stopReplay()` seam is added, used by the next slice to stop an in-flight recording at the source.

`sentrySink.test.ts` is rewritten from asserting "Session Replay not enabled" to asserting the ADR 0002 configuration as facts.

Replay capture is **not** yet live at this commit — the consent-at-source wiring (next slice) and the disclosure copy (slice after that) both land before capture can actually run, per the issue's ordering requirement.

Refs #166.

## Walkthrough

No UI change. This slice is Sentry SDK configuration only.

## Evidence

- `bun run typecheck` — pass
- `bun run check` — pass
- `bun run build` + `bun run verify:budget` — pass: 0 eager bytes in 6 chunks, 129,069 / 153,600 lazy bytes with replay configured (baseline on `main` was 128,917 — `replayIntegration` was already in the bundle via `@sentry/solidstart`'s barrel export, unused, before this stack)

## Acceptance criteria met

From #166:
- [x] `src/monitoring/sentrySink.test.ts` is updated from "Session Replay not enabled" to the ADR 0002 configuration as facts: masking on, media blocked, canvas off, `replaysOnErrorSampleRate` 0, session sampling low.
- [x] `bun run verify:budget` still shows zero SDK bytes before first paint, and the section 10 three-second-interactive budget is re-measured with replay enabled.

The remaining acceptance criteria (disclosure copy, consent-at-source stop, OPS-03 payload rule) are satisfied by the later slices in this stack.

---

### DIFF
--- a/src/monitoring/sentrySink.test.ts
+++ b/src/monitoring/sentrySink.test.ts
@@ -1,13 +1,19 @@
 // The Sentry SDK is never loaded here. Every test injects a fake module
 // through `SentrySinkOptions.load`, which is what lets this file assert the
-// ADR 0001 configuration decisions — `sendDefaultPii` off, console breadcrumbs
-// disabled rather than filtered, no Session Replay — as facts about the options
-// object rather than as comments nobody checks.
+// ADR 0001 and ADR 0002 configuration decisions — `sendDefaultPii` off, console
+// breadcrumbs disabled rather than filtered, Session Replay present with
+// masking on, media blocked, canvas off, and error-triggered replay at 0 — as
+// facts about the options object rather than as comments nobody checks.
 
 import { describe, expect, it, vi } from "vitest";
 import { UNKNOWN_BROWSER } from "./browserInfo";
 import type { ErrorReport } from "./errorReporting";
-import { SentrySink } from "./sentrySink";
+import { BLOCK_CONTENT, MASK_CONTENT } from "./replayPrivacy";
+import {
+	REPLAY_CANVAS_INTEGRATION,
+	REPLAY_SESSION_SAMPLE_RATE,
+	SentrySink,
+} from "./sentrySink";
 
 type InitOptions = Record<string, unknown>;
 
@@ -15,6 +21,7 @@ function fakeSentry() {
 	const inits: InitOptions[] = [];
 	const captures: { error: unknown; context: Record<string, unknown> }[] = [];
 	const close = vi.fn(async () => true);
+	const stopReplay = vi.fn(async () => {});
 	const integration = (name: string) => (options?: unknown) => ({
 		name,
 		options,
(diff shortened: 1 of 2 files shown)


afternoon:
@claude I commented on #190 that masking is overly aggressive. Masking text makes sense. Masking canvas or larger blocks of content, e.g. the mixer, is too aggressive.

afternoon:
## Answered: (b), canvas capture enabled — pushed as `1d50821`

You asked for canvas; I asked whether to take route (a) unmask-static-chrome or (b) canvas-on-with-an-ADR-amendment, and declined to make (b) from a PR comment. #210 is that amendment ([ADR 0003](https://github.com/afternoon/solid-groove/blob/claude/hard-006-adr-0003/docs/adr/0003-replay-canvas-capture-and-assistant-conversation-retention.md)), so this slice now takes (b).

**What changed here**

- `replayCanvasIntegration` is constructed alongside `replayIntegration` — both or neither, since canvas records nothing on its own and replay without it is the grey-box replay that made this rework necessary.
- The test asserting the canvas integration is *absent* now asserts it is **present**, with a second test pinning that an opted-out user constructs neither.
- Folded in the `maskAttributes` line flagged from #190 — that slice added `MASK_ATTRIBUTES` but had no replay integration to attach it to. The assertion pins the whole list, since the SDK replaces its default rather than extending it.
- `BLOCK_CONTENT` removed from the config, following its removal upstream. That was the compile break #190 predicted.

**On your point about canvas pixels being project content** — you were right, and the ADR says so in those terms rather than softening it. `canvasRenderer.ts:141` draws user-authored section names into the content layer, and masking has no reach inside a canvas. So this is all-or-nothing and it is deliberately "all": recordings include the arrangement, the notes, the waveforms, and those section names. The safeguard is not a technical measure that would not work — it is the corrected disclosure in #192 and the purpose limit in ADR 0003 decision 4.

Three scoping questions from #190 are settled in the ADR rather than left to the implementing PRs: `ConfirmDialog` masks its **title only** (your suggestion — landed in #190); asset/pad/pack names stay **unmasked**, revisited when user uploads land; `aria-label` masked via the explicit list.

**Evidence** — `bun run typecheck` clean, `bun run check` clean, 148 monitoring tests pass, 2161/2161 at the stack tip. The one red in the full run is `testAudioTeardown.test.ts`, which fails identically on a clean tree (`ConstantSourceNode`, no audio device) — confirmed pre-existing by stashing.

PR #192, opened 2026-08-06
HARD-006 - Disclose Session Replay before any capture is enabled (4 of 6)

afternoon:
## What & why

4 of 6 in the HARD-006 stack, builds on #191 (`claude/hard-006-replay-3`).

Updates `src/components/TelemetryDisclosure.tsx` copy to name Session Replay specifically and state, in plain language:

- that it records how the app is used — which controls are clicked, how the user moves through the app;
- that its purpose is to understand how people use Solid Groove so it can be improved;
- that it is **not** used to access the user's music or other private information;
- that project content is masked out at capture, so it never reaches Sentry at all.

`TelemetryDisclosure.test.tsx` asserts this copy is present, the same way it already covers the two existing processors.

This slice was deliberately inserted **before** the consent-at-source wiring slice (originally planned as part of the same branch as the wiring): had the wiring landed alone, replay capture would have gone live at `replaysSessionSampleRate` while the disclosure the user could read still named only two processors. With this slice landing first, the disclosure is always true at or before the moment capture becomes live at any commit in the stack.

Refs #166.

## Walkthrough

**Entrypoint:** the public landing page (`/`), dark theme, `bun run dev:mock`. The disclosure lives in the landing footer (`placement="inline"`, per `LOOP-001b`) — reachable without an account: scroll to the footer, click **Privacy**.

Verified by hand in the running app (reported by the implementer):
- The expanded panel shows all four statements ADR 0002 decision 5 requires — Session Replay named, its purpose ("so we can understand how people make music with Solid Groove and improve it"), what it is not for ("not used to access your music or anything else private"), and the masking ("your project content is masked out as the recording is made, so it never reaches Sentry at all").
- The pre-existing promise "Your music never leaves your project" is unchanged and now reads "Event reports, error reports, and replays carry no project, track, clip, or section names…".

## Evidence

- `bun run typecheck` — pass
- `bun run check` — pass

This branch passed an Opus review round in the implementation workflow (this slice was added during that round specifically to fix a disclosure-before-capture ordering finding — see issue #166 comments).

## Acceptance criteria met

From #166:
- [x] The disclosure names Session Replay, states its purpose (understanding usage), states that it is not used to access the user's music or private information, and states that project content is masked out. Covered by a test asserting the copy is present, as `TelemetryDisclosure.test.tsx` does today.

The remaining acceptance criteria (consent-at-source stop, OPS-03 payload rule) are satisfied by the later slices in this stack.

---

### DIFF
--- a/src/components/LandingPage.test.tsx
+++ b/src/components/LandingPage.test.tsx
@@ -267,7 +267,9 @@ describe("LandingPage (PRD PRJ-06)", () => {
 
 			await userEvent.click(screen.getByText("Privacy"));
 
-			expect(screen.getByText(/Google Analytics/)).toHaveTextContent(/Sentry/);
+			expect(screen.getByText(/Two processors receive this/)).toHaveTextContent(
+				/Sentry/,
+			);
 			expect(
 				screen.getByRole("checkbox", {
 					name: /share usage and error reports/i,
--- a/src/components/TelemetryDisclosure.test.tsx
+++ b/src/components/TelemetryDisclosure.test.tsx
@@ -17,7 +17,8 @@ function setup() {
 describe("TelemetryDisclosure (PRD OPS-02, section 10 Security and privacy)", () => {
 	it("names both processors, so the user is told what is collected", () => {
 		setup();
-		const body = screen.getByText(/Google Analytics/).textContent ?? "";
+		const body =
+			screen.getByText(/Two processors receive this/).textContent ?? "";
 		expect(body).toContain("Google Analytics");
 		expect(body).toContain("Sentry");
 	});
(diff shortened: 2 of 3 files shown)


afternoon:
## Rewritten for canvas-on — pushed as `4e3e21c`

The copy this slice shipped made three claims [ADR 0003](https://github.com/afternoon/solid-groove/pull/210) falsifies:

1. project content "is masked out as the recording is made, so it never reaches Sentry at all"
2. "Your music never leaves your project"
3. replays carry "no notes"

With canvas capture on, the arrangement and piano roll are recorded — clips, notes, and the section names the renderer draws into the canvas. All three were about to become false statements shown to users, which is a different order of problem from copy that reads awkwardly.

**What it says now:** recordings include the arrangement and piano roll, which is the musical work itself, and we record them because that is where the problems we need to see happen. Names and typed text still stay hidden — that claim is still true and still tested. Adds the assistant-conversation disclosure ADR 0003 decision 2 requires: stored with the project, never sent to Google Analytics or Sentry, deleted when the project is.

The test that pinned `Your music never leaves your project` now asserts that sentence is **absent**, so the superseded promise cannot return in a later copy edit.

## Walkthrough

Entrypoint: the public landing page (`/`), dark theme. Scroll to the footer, expand **Privacy**.

Disclosure copy

The four paragraphs as rendered:

> Solid Groove records which features are used and reports errors, so we can tell what works and fix what breaks. Two processors receive this: Google Analytics for product events and Sentry for error reports and Session Replay.
>
> Session Replay records a small sample of sessions — which controls you click and how you move around — so we can see where people get stuck and fix it. Those recordings include your arrangement and piano roll, which means the musical work itself: the clips, the notes, and the names you give sections. We record them because that is where the problems we need to see actually happen.
>
> Names and typed text stay hidden. Event reports, error reports, and replays carry no project, track, or clip names, no audio, no assistant messages, and no text you type.
>
> Your conversations with the assistant are stored with your project so we can tell whether it is helping. They stay with us — never sent to Google Analytics or Sentry — and deleting a project deletes its conversations.

Verified in the running app against the mock backend, not only in tests.

**Evidence** — `typecheck` clean, `check` clean, 37 tests pass across `TelemetryDisclosure.test.tsx` + `LandingPage.test.tsx`, 2161/2161 at the stack tip.

`DEC-009` still owns final wording; the constraint this changes is that the copy must be **true**, not that it must preserve a particular sentence.

Issue #226, opened 2026-08-06
Library browser reports "The library data could not be read" when the library has not been built


afternoon:
Opening the library shows a red "The library data could not be read." on a checkout where the library has not been rendered locally.

## What is happening

The client fetches the pack index same-origin:

```ts
// src/library/manifest.ts:35
export const LIBRARY_ROOT = "samples/starter-library";
export const PACK_INDEX_PATH = `${LIBRARY_ROOT}/packs/index.json`;

// src/library/libraryClient.ts:64
return fetch(`/${path}`).then(...)
```

That path is served out of `public/samples/starter-library/`, which `bun run library:build` renders and which is **gitignored** (`.gitignore:58`; `scripts/starter-library/build.mjs:35` — "Gitignored via `public/samples` — rendered audio is never committed"). A fresh clone has no `public/samples` at all — this one has only `favicon.ico` and `fixtures/`.

The missing file then does *not* 404. Both the dev server and Firebase Hosting rewrite unknown paths to the SPA shell (`firebase.json`: `"source": "**"` → `/index.html`), so the fetch returns **HTTP 200 with HTML**, `response.json()` throws, and the failure is classified `corrupt`:

```ts
// src/library/libraryClient.ts:71
try {
    return await response.json();
} catch {
    throw new FetchClassifiedError("corrupt", "Response was not JSON");
}
```

`corrupt` is labelled "The library data could not be read." (`src/library/loadReasons.ts`). So a *missing* library is reported as a *corrupt* one.

## Why it matters

Two distinct problems, either of which is worth fixing on its own:

1. **The library is absent by default.** Nothing tells a developer — or a deployed build — that a required delivery artefact was never produced. The app boots, the browser opens, and the failure looks like data corruption.
2. **The error is misleading, and `not_found` is unreachable.** The classifier's `not_found` branch (`libraryClient.ts:67`) can never fire for a library path behind a catch-all rewrite, so the one reason that would point at the real cause is the one reason that cannot be reported. Someone reading "could not be read" goes looking for corrupt JSON.

## Expected

- A missing library reports as missing, with an actionable message, rather than as corrupt. Distinguishing an HTML body from a JSON one at the fetch boundary is enough to tell the two apart even behind a rewrite.
- Whatever guarantees the library is present in a deployed build is explicit — either a build step that renders it, a preflight check that fails loudly, or documented setup. Today `bun run library:build` is required before the browser works and nothing says so.

## Notes

[...]

PR #231, opened 2026-08-06
Read the sample library from Cloud Storage rather than same-origin

afternoon:
## What & why

The library browser reported "The library data could not be read." on production. The pack index was never there to read.

`library:build` is not in the deploy chain (`deploy → predeploy → build → prebuild → samples`, which runs `generate-samples.mjs` and `emitRuntime.mjs` only), and `public/samples` is gitignored, so the deployed artefact carries just the three bootstrap assets. Reproduced locally with `bun run samples`:

```
public/samples/starter-library/audio/sha256/f1/66/f166266c….wav
public/samples/starter-library/audio/sha256/0a/d8/0ad83302….wav
public/samples/starter-library/audio/sha256/82/f8/82f89d1d….wav
packs/index.json present? -> NO
```

Hosting's `"source": "**"` → `/index.html` rewrite then answers the index fetch with 200 + the SPA shell, `response.json()` throws, and a *missing* library is reported as a *corrupt* one.

Publishing the library would not have fixed it either, which is the finding that shaped this PR: `upload.mjs` writes to Cloud Storage under `library/`, and the client only ever fetched same-origin (`libraryClient.ts:64`, `manifest.ts:35`). The two halves of the delivery pipeline never met. `docs/sample-library.md` §12 is explicit that the library lives in Cloud Storage, and `storage.rules` already opens `library/` to unauthenticated reads — only the client was missing.

So every library URL now resolves through one `libraryUrl()` against a configured bucket, falling back to same-origin when there is none:

- **Bucket configured** → `https://firebasestorage.googleapis.com/v0/b/<bucket>/o/<encoded library/…>?alt=media`. Firebase Storage's download endpoint, not `storage.googleapis.com`, because it is `storage.rules` — not GCS IAM — that makes `library/` public, and only this endpoint enforces those rules.
- **Unset** → `/samples/starter-library/…` exactly as before, so local development, the mock backend, and both browser suites keep serving `library:build`'s output out of `public/` unchanged.

The two layouts are identical below their root (build.mjs mirrors the bucket "exactly, so `upload.mjs` is a straight copy"), so a delivery key is the same string either way and only the origin differs. `libraryKey()` also normalizes the two forms the publishers write — root-relative from `build.mjs`, `library/…` object keys from `upload.mjs:202` — which the old `resolveManifestPath` did not handle.

**Configuration:** none new for production. `VITE_FIREBASE_STORAGE_BUCKET` is already a build input in both `ci.yml` and `preview.yml`. `.env.example` now documents that it feeds the client's library delivery as well as the SDK, and that `FIREBASE_STORAGE_BUCKET` (the uploader's) should match it.

**Deliberately out of scope**, per the product owner on #226:

- **Rendering the library in the deploy chain.** At ~42 MiB it belongs in the bucket, which is what this now reads. Tracked separately.
- **The `not_found` classifier.** A genuine 404 becomes reachable once the client fetches cross-origin, but the same-origin fallback keeps the catch-all-rewrite trap, so distinguishing an HTML body from a JSON one is still worth doing on its own. `Refs`, not `Closes`, for that reason.

**The bootstrap set is untouched.** `factoryLibrary.ts:109` builds its own same-origin URLs from `storageRef`, and those three files do ship in the Hosting artefact, so the starter project keeps working before any manifest fetch — §12's "keep only the bootstrap set in the initial application cache".

Refs #226

## Core flows

**Untouched.** #226 is a bug report and links no flow IDs. No spec was added, removed, or edited — `e2e/flows/CF-002.spec.ts` is the only flow file mentioning the library and it remains at `test.fixme` (it belongs to ARR-003, landed by #227). Verified: `git diff --stat origin/main -- e2e/ e2e-emulator/ docs/core-flows.md docs/prd.md` is empty.

Worth flagging for whoever picks up the follow-up: no e2e currently exercises the library browser against real delivery. `LibraryBrowser.tsx` → `useLibraryBrowser` → `new LibraryClient()` uses the real `httpJsonFetcher`, but every test injects a fetcher over `__fixtures__`. That is precisely why this shipped broken and stayed invisible.

## Walkthrough

**No UI change.** No component, layout, styling, copy, or interaction is touched — this changes only which origin the existing browser fetches from. The user-visible effect (the browser populating instead of showing the red error) cannot be captured here anyway: it needs the library published to the bucket, which is an ops step this PR does not perform.

## Evidence

```
$ bun run typecheck
$ tsc --noEmit                                    (clean)

$ bun run test
Test Files  167 passed (167)
     Tests  2255 passed (2255)

$ bun run check
Checked 483 files in 1074ms. Fixed 1 file.        (import ordering, committed)
```

Red before green — with the old `fetch(`/${url}`)` prefix restored, the new guard fails and everything else still passes:

```
× the default fetcher (issue #226) > requests the resolved delivery URL verbatim
✓ the default fetcher (issue #226) > classifies a 404 as not_found rather than corrupt
✓ lazy loading (sample-library section 12) > fetches the pack index and no manifest to open
  … 8 more passing
```

`bun run test:browser:chromium` — result appended as a comment when it finishes. Chromium-only is what this environment can install (it cannot reach `cdn.playwright.dev`), so it is a pre-flight, **not** PRD section 10 gating evidence; CI's full matrix on push is the cross-browser check.

Not run, with reasons: `test:emulator` and `test:browser:emulator` (no Firestore/Auth rules, persistence, or auth behavior touched); `library:test` (no script under `scripts/starter-library/` is modified — the bucket layout is read here, not changed).

## Acceptance criteria met

#226 has no acceptance checkboxes. Against its "Expected":

- ✅ **"Whatever guarantees the library is present in a deployed build is explicit."** It is delivered from the bucket `library:upload` publishes to, which is where §12 always said it lives, rather than depending on an undocumented local build step.
- ⬜ **"A missing library reports as missing, rather than as corrupt."** Not addressed — see out-of-scope above. Behind the same-origin fallback a missing library still reports `corrupt`.

One thing the PR cannot establish: I could not reach `https://groove-35c07.web.app` from my session (egress policy, 403 on CONNECT), so the production failure is diagnosed from the pipeline and reproduced locally, never observed live. Publishing the library to the bucket and confirming the browser populates is an ops step still outstanding — this change is what makes that publish take effect.

https://claude.ai/code/session_017soVbsGgTHzA8Ek8ueuBs4

---
_Generated by [Claude Code](https://claude.ai/code/session_017soVbsGgTHzA8Ek8ueuBs4)_

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,11 @@ VITE_FIREBASE_API_KEY=
 VITE_FIREBASE_AUTH_DOMAIN=
 VITE_FIREBASE_PROJECT_ID=
 VITE_FIREBASE_DATABASE_URL=
+# Also where the client reads the sample library from: the browser fetches the
+# pack index, pack manifests, and asset audio out of this bucket's `library/`
+# prefix (docs/sample-library.md section 12). Leave it unset and the library is
+# served same-origin from `public/samples/starter-library` instead, which is
+# what `bun run library:build` renders for local development.
 VITE_FIREBASE_STORAGE_BUCKET=
 VITE_FIREBASE_MESSAGING_SENDER_ID=
 VITE_FIREBASE_APP_ID=
@@ -42,7 +47,10 @@ VITE_FIREBASE_MEASUREMENT_ID=
 # bundle, and the service account belongs in CI secrets rather than on a
 # developer machine.
 #
-# Bucket to publish to. Falls back to VITE_FIREBASE_STORAGE_BUCKET above.
+# Bucket to publish to. Falls back to VITE_FIREBASE_STORAGE_BUCKET above, which
+# is also the bucket the client reads the library from — keep them the same
+# unless you are deliberately publishing somewhere else, or the browser will
+# read a library this never wrote to.
 # FIREBASE_STORAGE_BUCKET=your-project.firebasestorage.app
 #
 # Credentials — set exactly one:
--- a/src/library/libraryClient.test.ts
+++ b/src/library/libraryClient.test.ts
@@ -2,11 +2,16 @@ import { describe, expect, it, vi } from "vitest";
 import { fixtureFetcher } from "./__fixtures__/fixtures";
 import {
 	FetchClassifiedError,
+	httpJsonFetcher,
 	type JsonFetcher,
 	LibraryClient,
 	type LibraryIndexError,
 } from "./libraryClient";
-import { type LibraryPackSummary, PACK_INDEX_PATH } from "./manifest";
+import {
+	type LibraryPackSummary,
+	libraryUrl,
+	PACK_INDEX_PATH,
+} from "./manifest";
 
 async function loadedClient(): Promise<{
 	client: LibraryClient;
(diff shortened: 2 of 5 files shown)


afternoon:
Completing the evidence promised above — `bun run test:browser:chromium` finished green:

```
2 skipped
18 passed (34.5s)
```

The 2 skipped are `e2e/flows/CF-002.spec.ts`'s `test.fixme` markers, untouched by this PR (they belong to ARR-003, landed by #227).

Chromium-only, so this is a pre-flight rather than PRD section 10 gating evidence — this environment cannot reach `cdn.playwright.dev` to install Firefox or WebKit, and CI's full matrix on push is the cross-browser check.

---

Issue #228, opened 2026-08-06
Selecting a track in either mixer or arrangement view should show the instrument/devices for the track


afternoon:
Steps to repro:
 
- Open a new project
- Delete all tracks
- Create 2 new tracks

Expected:

1. I should be able to see the instruments/devices for a track
2. I should be able to switch between tracks by clicking them in either the mixer or the arrangement

Actual:

- I can't do either of the expected things

afternoon:
A three-PR stack fixes the selection half of this, and one part of it cannot be fixed here at all — flagging that rather than closing over it.

**Landing as:** #238 (view-model seam, no behavior change) → #239 (mixer selects; a clipless track shows its instrument) → #240 (arrangement selects; closes this issue). Split because the whole change is ~870 lines and the PR ceiling is 400.

**Both expectations from the report:**

- *Switch between tracks by clicking them in either the mixer or the arrangement* — done in both, by pointer and by keyboard. The edited track lives in the shared PRD 9.2 selection model, so the mixer strip and the arrangement header row always mark the same track, and a deleted track can't strand the editor.
- *See the instruments/devices for a track* — the **instrument**, yes, for whichever track is selected, including a track with no clip (which is every track added from the mixer; it previously showed nothing at all). **Devices, no.**

**On devices:** there is no device-chain UI anywhere in the editor. `src/audio/devices` has the DSP and `DeviceChain` reconciles an ordered chain onto every track, return, and the master, but nothing renders an insert chain for a track, so there is no panel to point at the selected track. That is a missing feature (PRD FX-01 / section 9.7), not a selection bug, and it seems worth its own issue rather than being folded into this one — happy to open it if you'd like. Whatever panel lands reads the same selected track this stack introduces.

Two neighbours this made obvious, both already filed: every track added from the mixer is a synth (#223), and there's no way to change it afterwards (#224). Selecting a track now shows you that synth, which is what makes those two the natural next fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

afternoon:
Filed the devices half as #241 — the missing device-chain panel, with what already exists below it (domain types, seven registered commands, the DSP, and the catalog keys) versus what does not (any UI at all, and therefore any `device_added` call site, which is LOOP-008's remaining unchecked criterion).

This issue stays scoped to selection, which #240 closes.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #238, opened 2026-08-06
Point the editor's derivations at a selected track (1 of 3, #228)

afternoon:
## What & why

1 of 3 for #228. Preparation only — **no behavior change**.

`editorViewModel` hard-coded the project's first track. `editedTrack` was
`song.tracks[0]`, and `editedClip`, `editedInstrument`, `sampleName`,
`showPianoRoll`, and `instrumentPanelTrackId` each re-derived it, so there was
nowhere to say "point the editor at *that* track" — the root of #228's "I can't
switch between tracks".

- `editedTrack(project, selectedTrackId)` resolves the selection, falling back
  to the project's first track when nothing is selected, or when the selected id
  names a track the project no longer has. That fallback is what keeps every
  derivation below it total, and keeps opening a project free of any selection
  gesture.
- The rest now take the resolved `Track` instead of re-deriving one.
- `drumTrack` follows the edited track rather than searching the project for the
  first drum-machine track: a project can hold several, and the one on screen
  should be the one you clicked.
- `focusedTrackId(selection)` reads the focused track out of the shared PRD 9.2
  `SelectionState` — where PR 2 holds the editor's selection. Selection stays
  UI-only state; nothing here touches a `Project`.

`EditorView` passes `null`, so it still edits the first track, exactly as
before. No domain, command, persistence, or audio contract is touched.

Refs #228

## Rebased onto #233/#235

This stack was written against `38c81ae` and rebased onto `efdad99` after #231,
#232, #233, #234 and #235 landed. One conflict here, resolved in `main`'s
favour: #235 had already widened `instrumentPanelTrackId` to name the edited
track whatever instrument it carries, so the kind picker can leave a drum
machine. That semantic is kept exactly; only the signature changes, from
`(project)` to the resolved `(track)`. Its test keeps `main`'s
drum-machine case and drops the "is null for a drum machine" case this branch
had written against the older behaviour — that assertion contradicted #235, and
belonged to it, not to this PR.

## Core flows

Untouched. This issue is a bug report and links no flow. `git diff origin/main
--stat -- e2e/ e2e-emulator/ docs/` is empty across this PR.

## Walkthrough

No UI change. Every rendered output is identical: the parameter added here is
passed `null` by the only caller. The walkthrough for the change users can see
is on the PR that closes the issue (3 of 3).

## Evidence

All re-run on the rebased commit (`9d96b92`), not carried over from before it:

- `bun run test` — `Test Files 170 passed (170)`, `Tests 2285 passed (2285)`,
  green on this commit on its own.
- `src/editor/editorViewModel.test.ts` grew by ten cases: the existing ones now
  pass the resolved track explicitly, plus new ones for the selected track, the
  two fallbacks (nothing selected, selected track gone), a project with no
  tracks, `focusedTrackId`, and `drumTrack` no longer answering for a track that
  is not the edited one.

## Acceptance criteria met

The issue has no checkboxes and this PR satisfies neither of its two symptoms on
its own — it is the seam the next two build on. Both are met by the end of the
stack: PR 2 (#239) makes the mixer select, PR 3 (#240) the arrangement.

### DIFF
--- a/src/editor/EditorView.tsx
+++ b/src/editor/EditorView.tsx
@@ -135,10 +135,13 @@ export default function EditorView(props: EditorViewProps): JSX.Element {
 		model.playheadLabel(audio.positionTicks()),
 	);
 
-	const track = createMemo(() => model.editedTrack(project()));
-	const drumTrack = createMemo(() => model.drumTrack(project()));
+	// The track the editor is pointed at. Selecting one lands with the mixer and
+	// arrangement wiring (#228); until then this is the project's first track,
+	// which is what `model.editedTrack` falls back to.
+	const track = createMemo(() => model.editedTrack(project(), null));
+	const drumTrack = createMemo(() => model.drumTrack(track()));
 	const sampleAssets = createMemo(() => model.sampleAssets(project()));
-	const clip = createMemo(() => model.editedClip(project()));
+	const clip = createMemo(() => model.editedClip(project(), track()));
 
 	// The step editor's live playback-step indicator (CLP-02): which 16th step of
 	// the edited clip the playhead is currently passing, wrapped within the clip's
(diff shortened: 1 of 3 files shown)


afternoon:
**CI on this PR: the unit job's failure was a flake; the Firefox one is the base's.**

- `typecheck, check:ci, unit + component tests` failed on
  `src/audio/devices/shaping.test.ts › adds harmonics as drive climbs`. I re-ran
  the failed jobs on the **identical commit** (`9d96b92`) and it passed, which
  is proof of non-determinism rather than a regression. This PR's diff is three
  editor files; nothing in that test's import graph. Filed with the evidence as
  **#251** — it also reproduces on this machine at roughly 2-in-4 in the full
  suite while passing in isolation, and `main` passes it 5-for-5.
- `browser e2e (firefox)` is the `CF-001` playback failure that fails identically
  on `main`; diagnosed once in [#240's thread](https://github.com/afternoon/solid-groove/pull/240#issuecomment-5209828743).

Nothing to fix here. The stack's one genuine CI failure was WebKit on #239,
which was mine and is fixed there — this PR is unaffected by it.

---

Issue #43, opened 2026-07-27
LOOP-003 - Transport, tempo, loop, and metronome


afternoon:
**LOOP-003 — Transport, tempo, loop, and metronome**

| Field | Value |
| --- | --- |
| Phase | 1 — complete the loop workflow |
| Dependencies | `FND-009` |
| PRD | AUD-01, AUD-02, AUD-03, AUD-04 |

Implement dependable play/pause/stop/seek, playhead, 40-240 BPM tempo, fixed 4/4 display, bar loop, metronome, master meter, and safety limiter.

## Acceptance criteria

- [x] Scheduling is ahead-of-time and remains aligned across seek, tempo change, repeated loops, and editor load.
- [ ] User-gesture unlock and focus-safe `Space` behavior work in all supported browsers. **Partially met.** Unlock is now bounded and fails loudly (`audio_start_failed`, `was_browser_blocked`) instead of hanging, and focus-safe `Space` is asserted unguarded in chromium, firefox and webkit. Playback itself still does not start in Firefox — measured, not assumed: `AudioContext.resume()` there stays pending forever. The cause is a real-browser-policy question, deferred to `HARD-001`'s cross-browser pass along with a user-facing "your browser blocked audio" state. See [`docs/testing.md`](../blob/main/docs/testing.md#playback-is-asserted-in-chromium-only--a-known-tracked-gap).
- [x] Master safety and meter tests include silence, extreme chains, and no transport restart on parameter edit.
- [x] Emits `transport_play` (with `surface` and `is_first_play_in_session`), `audio_start_failed` with a browser-blocked flag when the context cannot unlock, and sampled `audio_underrun` events with the sampling rate recorded. Playback emits nothing per scheduled event or animation frame.

**Deferred, deliberately:** AUD-02's user-defined loop *range*. The engine side (`barAlignedLoop`, `TransportController.setLoop`) is built and tested, but this slice renders a 16-step grid and no timeline, so there is nothing to select a range on — the loop is the one bar the grid shows. The range-selection UI belongs with `ARR-002`, which owns range select and range loop.

This issue is the live record for LOOP-003: it is the specification as well as the status. Status is its state and labels, ownership is its assignee, and progress, blockers, and evidence are comments. Product behavior is authoritative in [`docs/prd.md`](../blob/main/docs/prd.md).

afternoon:
## Finding from `FND-009`: a blocked Firefox unlock hangs silently and emits no `audio_start_failed`

[...]

PR #239, opened 2026-08-06
Follow the track you select in the mixer (2 of 3, #228)

afternoon:
## What & why

2 of 3 for #228, builds on #238.

The editor showed the project's first track and nothing else. Delete the tracks,
add two, click either one: the same instrument stayed on screen and the other
was unreachable.

- **The edited track lives in the shared PRD 9.2 selection model.**
  `EditorView` owns a `SelectionState` — UI-only, never persisted, never in a
  `Project` — and a click on a strip replaces it with that track's scope. The
  clip editor, the instrument panel, the kind picker, and the drum-machine pads
  all move together. Nothing is selected on arrival, which still resolves to the
  first track (#238's fallback).
- **A deleted track can't strand the editor.** A `reconcileSelection` pass on
  every project change drops a scope whose track a delete, an undo, or a remote
  edit removed, and the fallback picks up from there.
- **The whole strip is the target.** `pointerdown` anywhere on it selects —
  reaching for a fader *is* working on that track, and doing it on `pointerdown`
  means the panel you are about to adjust is already the one on screen. The
  colour chip became an `Edit <track>` button so a keyboard has one labelled
  control per strip. It is named "Edit", not "Select", because the arrangement's
  accessible list already owns `Select <track>` for selecting a bar range and
  `CF-002` clicks it by that name; two identically named controls would make
  that locator — and a screen reader — ambiguous.
- **Adding a track selects it**, since the reason to add one is to set it up.
- **A clipless track renders its instrument panel**, with one line where the
  grid would be. A project with no tracks at all says so and points at the
  mixer.

Selection emits no event of its own: it reports `feature_first_use: mixer`, the
key every other strip interaction already reports. It is a pointer landing in
the mixer, not a new tracked action, and no PRD OPS-02 event covers "changed
which track is in focus".

Arrangement clicks are PR 3. No domain, command, persistence, or audio contract
is touched.

Refs #228

## Rebased onto #233/#235

Written against `38c81ae`, rebased onto `efdad99`. Two conflicts, both resolved
in `main`'s favour with this branch's change re-applied on top:

- **`Mixer.tsx`** — #233 replaced the single "Add track" with one button per
  instrument kind, through `createNewTrack`. The "select what you just added"
  line now follows that path (`added.track.id`) and applies to all three
  buttons; its test clicks "Add sampler track". Duplicating a track
  deliberately still does *not* move the selection — that would be a second
  behaviour this PR does not describe.
- **`TrackEditor.tsx`** — #235 added the `InstrumentKindPicker` above the
  instrument panel. It stays *outside* the clip `Show` this PR introduces, so a
  track with no clip still gets the picker as well as the panel. That is the
  same intent from both sides, not a compromise.

## Core flows

Untouched — no flow spec's assertions changed; `git diff origin/main --stat --
e2e/flows/ e2e-emulator/ docs/` is empty.

One non-flow browser spec did change, and it is a copy change, not a weakened
assertion: `e2e/smoke.spec.ts`'s "creates a genuinely empty project via Blank
Project" asserted the editor says *"This project has no sampler track yet."* A
blank project has no tracks of any kind, and since the editor now follows a
selected track of any kind, that sentence was describing the wrong absence. It
asserts the new text — *"This project has no tracks yet. Add one in the
mixer."* — and still proves the same thing.

## Walkthrough

Deferred to 3 of 3, which closes the issue and carries the capture.

## Evidence

All re-run on the rebased commit (`5f044f0`):

- `bun run test` — `Test Files 170 passed (170)`, `Tests 2291 passed (2291)`,
  green on this commit on its own.
- Chromium browser suite was run on the stack tip (see #240); this slice's own
  browser-visible change is the empty-project copy, covered there.
- Diff is 406 changed lines, six over the 400 ceiling — all six from the
  rebase's re-indentation in `Mixer.test.tsx` and `TrackEditor.tsx`, not new
  work. Flagging rather than shaving a test to get under it.

New tests:

- `Mixer.test.tsx` — `Mixer track selection (#228)`: selecting from a strip
  reports the right track and marks it (`aria-pressed`, `.selected`);
  `pointerdown` on a fader selects before it moves; selection logs exactly one
  `feature_first_use: mixer` and nothing else. Plus, under TRK-01, that adding a
  track selects it.
- `EditorView.test.tsx` — a clipless track shows its instrument panel and
  neither clip editor; clicking a mixer strip swaps the editor between the drum
  fixture's two tracks and back.

## Acceptance criteria met

The issue has no checkboxes. Of its two symptoms:

- [x] "I should be able to see the instruments/devices for a track" — for the
      instrument, including on a track with no clip. **Devices: not possible**,
      and now tracked as #241 — there is no device-chain UI anywhere in the
      editor to point at a track. The hook is here; the panel does not exist.
- [ ] "I should be able to switch between tracks by clicking them in either the
      mixer or the arrangement" — mixer only, here. The arrangement is #240.

### DIFF
--- a/e2e/smoke.spec.ts
+++ b/e2e/smoke.spec.ts
@@ -219,8 +219,12 @@ test.describe("dashboard project management", () => {
 		await page.getByRole("button", { name: "Blank Project" }).click();
 
 		await expect(page).toHaveURL(/\/projects\/prj_/);
+		// A blank project has no tracks at all. The editor says so and points at
+		// the mixer, rather than reporting the absence of a sampler track — since
+		// #228 the editor follows a selected track of any kind, so "no sampler
+		// track" was never what an empty project was short of.
 		await expect(
-			page.getByText("This project has no sampler track yet."),
+			page.getByText("This project has no tracks yet. Add one in the mixer."),
 		).toBeVisible();
 	});
 
--- a/src/editor/EditorView.css
+++ b/src/editor/EditorView.css
@@ -245,6 +245,16 @@
 	color: var(--color-text-secondary);
 }
 
+/* Where the step editor or piano roll would be for a track that has no clip
+   yet. Set in the panel's own body text rather than shouted, because the track
+   is fine — there is simply nothing to program on it. */
+.track-editor .no-clip {
+	margin: 0;
+	padding: 12px 0;
+	font-size: var(--font-size-small);
+	color: var(--color-text-secondary);
+}
+
 /*
  * The library panel (LOOP-013). Open by default and closed from the header
  * toggle, it is a narrow left-hand column running the full height of the editor
--- a/src/editor/EditorView.test.tsx
+++ b/src/editor/EditorView.test.tsx
@@ -76,6 +76,14 @@ function saveStatusGroup(): HTMLElement {
 	return group;
 }
 
+/** The mixer's "edit this track" control for one strip. */
+function mixerSelect(trackName: string): HTMLElement {
+	return within(screen.getByRole("region", { name: "Mixer" })).getByRole(
+		"button",
+		{ name: `Edit ${trackName}` },
+	);
+}
+
 /** Null when the save state offers no retry: non-retryable, or not failed. */
 function saveRetryButton(): HTMLElement | null {
 	return within(saveStatusGroup()).queryByRole("button", { name: "Retry" });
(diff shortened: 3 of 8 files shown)


afternoon:
**CI: `browser e2e (firefox)` red — same base-branch failure as the rest of the stack.**

Checked, not assumed: this job fails on `CF-001 › a visitor with no account
reaches a playing loop` at `e2e/flows/CF-001.spec.ts:95` (`Stop playback` never
appears), `1 failed / 2 skipped / 17 passed`. `main`'s own run at `38c81ae`
([31128968506](https://github.com/afternoon/solid-groove/actions/runs/31128968506))
fails on the identical spec, line and counts, so it predates this branch. Full
diagnosis and the `docs/testing.md` / #43 background are in
[#240's thread](https://github.com/afternoon/solid-groove/pull/240#issuecomment-5209828743)
rather than repeated here.

The run measured `f33c40d`, which is no longer the head: the stack has been
rebased onto `efdad99` and this branch is now `5f044f0`. Nothing in this slice
touches the transport or `CF-001`.

---

afternoon:
**WebKit was mine. Fixed and pushed (`34b38da`).**

`browser e2e (webkit)` failed here on `e2e/mixer.spec.ts › creates a track of
the instrument the button names`: after clicking "Delete Sampler", the
confirmation dialog never appeared. `main` at the same base (`efdad99`) is
**green** in WebKit — [run 31129974187](https://github.com/afternoon/solid-groove/actions/runs/31129974187),
where only Firefox fails — so unlike the Firefox red, this one was not
inherited.

**Cause.** The whole-strip selection this branch added. It had to fire on
`pointerdown` to beat a fader drag, and WebKit only dispatches a `click` when
mousedown and mouseup land on the same element — so any re-render triggered
from that handler costs whatever control was actually pressed its own click.
Chromium and Firefox are more forgiving, which is why only WebKit caught it,
and why my Chromium pre-flight went green twice while this was broken.

**Fix.** Selecting is now the `Edit <track>` button beside the name and nothing
else: a real focusable control, on `click`, with nothing underneath it to break.
It is chip-sized plus a pointer-sized margin, and takes the accent when its
track is the one on screen. The strip container is inert again — which also
drops the two accessibility rules a clickable `div` had to suppress
(`useKeyWithClickEvents`, `noStaticElementInteractions`). Silencing those to keep
a nicety that had just caused a cross-browser defect was the wrong trade.

**What it costs**, stated plainly: touching a fader no longer selects that track
on the way past. The drag still applies; bringing the panel with it is now one
deliberate click. `Mixer.test.tsx` keeps the regression — pressing delete opens
the confirmation and selects nothing.

**Honesty about the diagnosis:** I could not reproduce this locally. WebKit's
binary cannot be fetched from this environment, and the same sequence passes in
jsdom (I wrote the repro) and in Chromium. The diagnosis is from the failure's
shape plus the main-vs-branch comparison. If WebKit stays red after this, the
cause is elsewhere and I will say so rather than keep guessing.

Re-verified on the new commit: `bun run test` — 2291 passed; `bun run check` and
`typecheck` clean; `bun run test:browser:chromium` — 20 passed.

---

Issue #223, opened 2026-08-06
Add track always creates a synth — no way to create a sampler or drum machine track


afternoon:
The mixer's **Add track** button is the only way to create a track, and it always mints a synth:

```ts
// src/editor/Mixer.tsx:564
function createTrackRecord(order: number): Track {
	return createTrack(factoryContext, {
		name: `Track ${order + 1}`,
		order,
		type: "instrument",
		instrument: createSynthInstrument(),
	});
}
```

So a user cannot create a sampler track or a drum machine track at all, even though both instruments are implemented (`src/audio/instruments/sampler.ts`, `drumMachine.ts`), have panels (`src/instrument/SamplerPanel.tsx`, `src/editor/DrumMachinePanel.tsx`), and are reachable from the domain (`createSamplerInstrument`, `createDrumMachineInstrument`).

## Why it matters

The starter project arrives with one sampler track (`src/editor/starterProject.ts`), so the only sampler a user ever gets is the one they were given. Building a multi-part loop — the Alpha Milestone 1 exit criterion, "an original 1-8 bar multi-track electronic loop ... including a drum machine and multiple chained effects" — requires adding sampler and drum machine tracks by hand.

It also blocks the ARR-003 core flow (#61), which builds a five-part loop before triggering the loop-to-song action.

## Expected

Creating a track lets the user say what kind it is — sampler, drum machine, or synth — and the new track opens on the matching editor (a synth note clip gets the piano roll; a sampler's gets the step grid).

## Notes

Track creation already goes through the `track.add` command, so this is a UI affordance over an existing command rather than a new mutation path. `track_added` is already emitted with `track_type` (`Mixer.tsx:122`), which today can only ever report `instrument`.

PR #240, opened 2026-08-06
Select a track by clicking it in the arrangement (3 of 3, #228)

### LINKED ISSUE #228: Selecting a track in either mixer or arrangement view should show the instrument/devices for the track

afternoon:
## What & why

3 of 3 for #228, builds on #239.

The mixer can now point the editor at a track. The arrangement — the surface you
are actually looking at while you work — could not, which is half of what the
issue asked for.

It takes the same two optional props the mixer took, and holds no
selected-track state of its own: it reports the track and `EditorView`'s
selection decides. That is what keeps the mixer strip and the header row marking
the same track rather than drifting apart.

- **Each windowed track header row is an `Edit <track>` button** with
  `aria-pressed`, so the header column *is* the control rather than a label
  beside one, and a keyboard reaches every visible track by tabbing it. The
  selected row takes the same accent edge as the mixer's selected strip. Named
  "Edit" for the reason given in #239: `Select <track>` already belongs to the
  accessible list's bar-range control below, which `CF-002` clicks by name.
- **A pointer landing anywhere on a row in the timeline selects that row's
  track**, before the click goes on to do whatever it was going to do — start a
  placement drag, or select a bar range. Either way the work is on that track.
- **The accessible track list does both**: it still selects the track's first
  bar range (unchanged, as `ArrangementView.test.tsx` asserts) and now also
  points the editor at it.

Selecting counts as `feature_first_use: arrangement`, like every other
interaction on this surface, and emits no event of its own. No domain, command,
persistence, or audio contract is touched; the ARR-001 shell, the ARR-002
placement controller, and the canvas renderer are untouched.

Closes #228

## Rebased onto #233/#235

The stack was written against `38c81ae` and rebased onto `efdad99` after #231 –
#235 landed; #238 had gone un-mergeable. This slice rebased without conflicts —
the two that needed resolving are in #238 and #239 and are described there. The
walkthrough below was re-captured after the rebase, so it shows the editor as it
actually ships, kind picker and all.

## Core flows

Untouched. `git diff origin/main --stat -- e2e/flows/ e2e-emulator/ docs/` is
empty across the whole stack; this issue links no flow of its own. (The one
non-flow spec the stack edits is `e2e/smoke.spec.ts`, in #239, for the reason
given there.)

### CF-001 — A visitor with no account reaches a playing loop

**1. Open the landing page**

CF-001 step 1

**2. You arrive at the dashboard as a guest, with no projects yet**

CF-001 step 2``

**3. The new project opens on a step editor with a starter pattern**

CF-001 step 3``

**4. Turn on a step that was off**

CF-001 step 4

**5. Start playback — the transport is running**

CF-001 step 5

<sub>Captured by `bun run walkthrough:capture` from the core-flow specs, in Chromium. Flows are defined in [`docs/core-flows.md`](https://github.com/afternoon/solid-groove/blob/main/docs/core-flows.md).</sub>

Step 3 shows the arrangement's "BD" header row carrying the selected marking —
the accent edge on the track the editor is showing. **It does not show the fix
itself**, and that is worth saying plainly rather than implying otherwise: every
registered flow works a one-track project, so none of them clicks a second
track, and a walkthrough is a byproduct of the flows rather than something
assembled by hand. What proves the switch is `EditorView.test.tsx`'s two
end-to-end cases — mixer strip and arrangement header, each swapping the editor
between the drum fixture's two tracks. A flow that adds a second track and
switches to it would be worth registering; that is the product owner's call, not
something to add to `docs/core-flows.md` from here.

## Evidence

All re-run on the rebased commit (`c9adec0`), not carried over from before it:

- `bun run test` — `Test Files 170 passed (170)`, `Tests 2296 passed (2296)`.
- `bun run test:browser:chromium` — `20 passed`, `CF-001` included. Chromium
  only: this environment cannot reach `cdn.playwright.dev` for the
  Firefox/WebKit binaries, so those are CI's to gate. This is **not** the PRD
  section 10 cross-browser evidence.
- `bun run verify:core-flows` — `3 flow(s), each with exactly one spec`, with
  `CF-002`/`CF-003` still parked at `test.fixme` (ARR-003's, untouched here).
- Diff is 232 changed lines.

Per-slice greenness (stacks item (a)): the full unit suite was run on each of
the three rebased commits in turn, not only at the tip — 2285, 2291, 2296 tests,
all green.

**CI's `browser e2e (firefox)` is red, and it is the base branch's failure.**
`CF-001` cannot start playback on a Firefox runner with no audio device; `main`'s
own run at `38c81ae` fails on the same spec, same line, same counts. Diagnosed
with the evidence in [this comment](https://github.com/afternoon/solid-groove/pull/240#issuecomment-5209828743);
it is the gap `docs/testing.md` documents and #43 tracks.

New tests in `ArrangementView.test.tsx` (`arrangement track selection (#228)`):
selecting from a header row reports the track and marks it; a pointer on a
timeline row reports that row's track; the accessible list reports the track
*and* keeps its bar-range selection; and selection logs exactly one
`feature_first_use: arrangement` and nothing else. Plus the end-to-end case in
`EditorView.test.tsx`, which also asserts the mixer marks the track the
arrangement selected.

## Acceptance criteria met

The issue has no checkboxes. Of its two expectations:

- [x] "I should be able to switch between tracks by clicking them in either the
      mixer or the arrangement" — both, by pointer and by keyboard.
- [x] "I should be able to see the instruments/devices for a track" — the
      **instrument**, for whichever track is selected, including a track with no
      clip (#239), and since #235 landed, its kind picker too.
- [ ] **Devices: still not possible, and not fixable here** — now filed as
      **#241**. The editor has no device-chain UI at all: `src/audio/devices`
      holds the DSP, `DeviceChain` reconciles it, seven `device.*` commands are
      registered, and `device_added` is in the analytics catalog — but nothing
      renders an insert chain, so there is no panel to point at the selected
      track, and `device_added` has never had a call site. That is a missing
      feature (PRD FX-01), not a selection bug. When it lands it reads the same
      selected track this stack introduces.

Two neighbours this stack noticed have since landed on `main` and are now part
of what it rebased onto: choosing a new track's instrument (#223 → #233) and
changing it afterwards (#235). Selecting a track is what makes both reachable
for any track but the first.

### DIFF
--- a/src/arrangement/ArrangementView.css
+++ b/src/arrangement/ArrangementView.css
@@ -88,11 +88,36 @@
 .arrangement-header-row {
 	display: flex;
 	align-items: center;
-	gap: 8px;
-	padding: 0 8px;
 	font-size: 12px;
 	box-sizing: border-box;
 	border-bottom: 1px solid #1e1e1e;
+}
+
+/* The whole row is the select control (#228): it fills the header cell and
+   carries no chrome of its own, so the column still reads as a list of tracks
+   rather than a column of buttons. The row whose track the editor is showing
+   takes the accent edge the mixer's selected strip does. */
+.arrangement-header-select {
+	display: flex;
+	align-items: center;
+	gap: 8px;
+	width: 100%;
+	height: 100%;
+	padding: 0 8px;
+	border-radius: 0;
+	font: inherit;
+	text-align: left;
+	background-color: transparent;
+	color: inherit;
+}
+
+.arrangement-header-select:hover {
+	background-color: #232323;
+}
+
+.arrangement-header-select[aria-pressed="true"] {
+	background-color: #232323;
+	box-shadow: inset 2px 0 0 var(--color-accent, #20c8e8);
 }
 
 .arrangement-header-swatch {
--- a/src/arrangement/ArrangementView.test.tsx
+++ b/src/arrangement/ArrangementView.test.tsx
@@ -1,4 +1,11 @@
-import { cleanup, fireEvent, render, screen } from "@solidjs/testing-library";
+import {
+	cleanup,
+	fireEvent,
+	render,
+	screen,
+	within,
+} from "@solidjs/testing-library";
+import { createSignal } from "solid-js";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import { Analytics } from "../analytics/analytics";
 import { ConsentStore } from "../analytics/consent";
@@ -7,6 +14,7 @@ import {
 	createLargeArrangementProject,
 	createSliceFixtureProject,
 } from "../domain/fixtures";
+import type { TrackId } from "../domain/ids";
 import { TICKS_PER_BAR } from "../domain/time";
 import { EditorSession } from "../editor/EditorSession";
 import { createInMemoryProjectRepository } from "../persistence/inMemoryProjectRepository";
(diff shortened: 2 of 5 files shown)


afternoon:
**CI: `browser e2e (firefox)` red, and it is the base branch's, not this stack's.**

The failure is `CF-001 › a visitor with no account reaches a playing loop`, at
`e2e/flows/CF-001.spec.ts:95` — after clicking play, `Stop playback` never
appears. Three retries, same result; the other 17 Firefox tests pass.

`main`'s own push run at `38c81ae`
([31128968506](https://github.com/afternoon/solid-groove/actions/runs/31128968506))
fails identically: same spec, same line, same locator, same `1 failed / 2
skipped / 17 passed`. So it predates this branch and reproduces without it.

It is also the documented gap rather than a surprise — [`docs/testing.md`](https://github.com/afternoon/solid-groove/blob/main/docs/testing.md#playback-is-asserted-in-chromium-only--a-known-tracked-gap)
records that playback is asserted in Chromium only, because a runner with no
`/dev/snd` leaves Firefox's `AudioContext` suspended, and `LOOP-003` (#43) holds
the product-side fix. Nothing in this stack touches the transport, the audio
runtime, or `CF-001`'s assertions.

Two notes on what those runs actually measured:

- They ran against `1f60ab6`, which no longer exists on the branch. The stack
  has since been rebased onto `efdad99` (#233/#235 had landed and #238 was
  conflicting), so the head is now `c9adec0` and fresh runs should follow.
- Chromium is green locally on the rebased tip: `bun run test:browser:chromium`
  → `20 passed`, including `CF-001`. Firefox and WebKit binaries cannot be
  fetched from this environment, which is why CI is the one that gates them.

No action taken on this failure. If the base ever goes green in Firefox and this
branch does not, that is mine and I will fix it.

---

Issue #225, opened 2026-08-06
A library sound cannot be loaded onto a sampler — drag from the browser to the instrument


afternoon:
A user can browse and audition the whole library but cannot put any of it on a track.

Two halves are missing:

**The browser's insert action is never rendered.** `AssetRow` draws its "Insert" button only when an `onInsert` handler is supplied (`src/library/AssetRow.tsx:61-70`), and `EditorView` mounts `LibraryBrowser` without one (`src/editor/EditorView.tsx:281-292`) — it passes `previewEngine`, `addedPackIds`, `onAddPack`, and `onPackBrowserOpenChange` only. The prop is threaded all the way down through `LibraryBrowser` and is simply never provided.

**The sampler's swap list cannot reach new content.** `replacementOptions` returns only the assets the project already carries:

```ts
// src/editor/editorViewModel.ts:128
/**
 * Samples the sampler panel can swap to: every `sample`-kind asset the
 * project already carries. A richer, genre-filtered browser lands with
 * LOOP-013; here it is a straight list of the project's own samples.
 */
```

LOOP-013 (#54) shipped that browser, but the comment's promised follow-through never happened, so the swap list is still limited to sounds already in the project — which, for a new project, means the single starter sample.

## Why it matters

The library is 200 sounds a user can hear and cannot use. It makes the browser a demo rather than a tool, and it blocks any multi-part loop, including the ARR-003 core flow (#61) and the Alpha Milestone 1 exit criterion.

LOOP-013's own scope line says "pack-organized browsing, search, filters, keyboard navigation, sync-aware audition, **insertion**, loading/error states, and clean preview teardown" — insertion is named there, and none of that issue's acceptance criteria cover it, which is how it closed without the wiring.

## Expected

Dragging an asset from the library onto a sampler instrument loads it, dispatching the existing `instrument.setSample` command so it is one undoable transaction and emits `instrument_changed`.

Per the product owner: the swap list can be dropped in favour of a simple label showing the loaded sample's name, since drag-and-drop replaces it as the way to choose a sound.

## Notes

Drag-and-drop needs a keyboard-reachable equivalent to satisfy PRD 9.3's rule that a pointer gesture is never the only way to drive a surface — the existing `onInsert` path is the obvious candidate, inserting onto the focused track.

Issue #61, opened 2026-07-27
ARR-003 - Sections and loop-to-song workflow


afternoon:
**ARR-003 — Sections and loop-to-song workflow**

| Field | Value |
| --- | --- |
| Phase | 2 — arrangement and export |
| Dependencies | `ARR-002`; #223, #225 |
| PRD | ARR-02, ARR-03; OPS-02 |
| Core flows | CF-002, CF-003 |

Implement song section markers on the arrangement timeline — add, rename, resize, reorder, color — and the loop-to-song action that turns a selected loop range into a named, editable arrangement outline from a structure template.

**The outline makes no musical decisions, by design.** It stamps a labelled canvas: named, coloured sections of varying length, each carrying a copy of every track in the selected loop. Which elements play where is the producer's to decide by subtraction afterwards, and the assistant's to propose in Alpha Milestone 3 — whose exit criterion is turning a reference loop into a valid editable outline with real project analysis behind it. A template that guessed at density (drop the drums in the intro) would need per-track role metadata the domain does not have, and would be a worse duplicate of that later feature. Section lengths vary so the result has shape rather than being N identical blocks.

## Acceptance criteria

- [ ] Sections are added, renamed, resized, reordered, and colored through registered commands, so a template or the assistant creates them by the same path the UI does. A marker labels a range on the shared timeline and holds no audio of its own.
- [ ] Reordering a section moves its contained placements as one undoable transaction — one revision, one history entry — leaving source clips unmodified.
- [ ] Loop-to-song offers at least one editable structure template: selecting a loop range and applying the template duplicates placements into named sections without destructively modifying source clips, the result is immediately playable, and one undo reverts the whole outline atomically.
- [ ] Emits `section_created` with `origin`, `arrangement_outline_created` with `template_id` and `section_count`, and `arrangement_milestone` when a project first reaches ≥3 named sections and ≥2 minutes, plus the `sections` `feature_first_use` key. This task claims the `template_id` keys, which are `UNCLAIMED` in `src/analytics/catalog.ts` today. Resize and reorder drags emit nothing per frame — one event per completed action.
- [ ] CF-002 and CF-003 pass with their `test.fixme` markers removed, and the closing PR carries the captured walkthrough.

## Core flows

[...]

PR #242, opened 2026-08-06
Add asset.add and asset.remove to the command registry

afternoon:
## What & why

1 of 5 for #225 — the central-registration slice, landing first so a sibling
feature cannot collide with it (CLAUDE.md "Landing work" item 3).

A project stores a *reference* to a library sound, not the library record, so
putting a sound on a track is two things: the project has to carry the asset,
and something has to point at it. The registry had the second half only —
`instrument.setSample`, `drum.setPadAsset`, an audio-loop clip — and no way at
all to carry a new asset. That is the command-layer reason the library browser
can audition 200 sounds and load none of them.

`asset.add` carries one, `asset.remove` drops one. Neither touches
`metadata.packDependencies` or the pack shelf: `executeTransaction` already
recomputes both from `song.assets` once per transaction, so the first sound
drawn from a pack shelves that pack atomically with the edit that needed it
(PRD LIB-05, invariant 12).

Two refusals, rather than a transaction that fails later as an invalid project:

- adding an asset the project already carries;
- removing one a track's instrument, a drum pad, or an audio-loop clip still
  resolves — named in the message, so it says what to clear first.

Removal's inverse restores the asset at the index it came from, because
`song.assets` is serialized in order and `inverse.test.ts` compares the
serialized content.

Registration points touched, and nothing else: `src/commands/registry.ts`,
`src/commands/index.ts`, and `COMMAND_IDS` in `src/analytics/catalog.ts`
(`catalog.test.ts` asserts that list equals the registry exactly).

The command test fixture grows one carried-but-unloaded asset, from a pack it
already depends on, because a removal test needs a target nothing holds.

Refs #225

## Core flows

Untouched. This is a mid-stack slice with no user-visible surface. #225 is a bug
with no core flows of its own; CF-002 (#61) depends on it and stays `test.fixme`
until #223 also lands — see #246.

## Walkthrough

No UI change.

## Evidence

```
bun run typecheck   → clean
bun run check       → Checked 502 files. No fixes applied.
bun run test        → Test Files 170 passed (170) · Tests 2278 passed (2278)
```

Diff size: 4 product lines + registrations + tests, well inside the 400-line
ceiling.

`inverse.test.ts` covers both new commands (it fails if a registered command has
no inverse case), and `registry.test.ts`'s pinned command list is updated
deliberately rather than derived.

## Acceptance criteria met

#225 has no acceptance checkboxes. This PR satisfies the command-layer
precondition of its "Expected" outcome: loading a sound is one undoable
transaction. The user-visible half lands in PRs 3–5 of this stack.

## Stack

1. **this PR** — `asset.add` / `asset.remove` (base `main`)
2. #243 — map a browsed library sound onto a project asset
3. #244 — let a library row carry its sound on a drag
4. #245 — load a dropped sound onto a sampler

### DIFF
--- a/src/analytics/catalog.ts
+++ b/src/analytics/catalog.ts
@@ -240,6 +240,8 @@ export const COMMAND_IDS = [
 	"instrument.setSample",
 	"pack.add",
 	"pack.remove",
+	"asset.add",
+	"asset.remove",
 	"device.add",
 	"device.remove",
 	"device.reorder",
--- a/src/commands/assets.test.ts
+++ b/src/commands/assets.test.ts
@@ -1,0 +1,187 @@
+import { beforeEach, describe, expect, it } from "vitest";
+import { createAsset, createPack, type Project } from "../domain";
+import { addAsset, removeAsset, setPadAsset, setSample } from ".";
+import { executeCommand, executeTransaction } from "./execute";
+import { CommandHistory } from "./history";
+import { findTrack } from "./projectEdits";
+import {
+	type CommandTestProject,
+	createCommandTestProject,
+	createTestFactoryContext,
+} from "./testProjects";
+
+/**
+ * `asset.add` / `asset.remove` (LIB-05, invariant 12).
+ *
+ * The behaviour worth pinning is not the array splice: it is that carrying a
+ * sound and pointing at it commit as one transaction, that the derived pack
+ * dependency list follows the assets rather than being maintained alongside
+ * them, and that an asset something still resolves cannot be dropped.
+ */
+
+function apply(
+	project: Project,
+	command: Parameters<typeof executeCommand>[1],
+): Project {
+	const result = executeCommand(project, command);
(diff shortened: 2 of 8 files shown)


afternoon:
CI red on `browser e2e (firefox)` — pre-existing on `main`, not this PR. This slice is the command registry plus tests; it renders nothing.

1 failed job of 8, `e2e/flows/CF-001.spec.ts:95` (the transport's "Stop playback" never appears in Firefox). `main` at `efdad99e` fails on exactly the same single job.

Raised as #249, with the evidence and a suggested fix — CF-001 asserts playback unguarded while the two other specs that press play guard it with `browserName === "chromium"`, so `main` and every branch cut from it are red. Not fixing it from here: it changes a frozen flow spec's assertions.

Everything this PR is responsible for is green — typecheck, lint, unit suite (2278 passing), chromium and webkit e2e.


---

PR #220, opened 2026-08-06
Adopt a core-flow-first workflow for implementing features

afternoon:
## What & why

Implements the revised feature workflow we agreed after reviewing the proposal. Features are now specified as **core flows** — plain-English user journeys with stable `CF-` ids, registered in `docs/core-flows.md` by the product owner — and a feature's first PR turns each linked flow into a Playwright spec that is reviewed *before* any implementation exists.

Six decisions are encoded here:

1. **Flow specs land as `test.fixme`, not red.** A deliberately failing PR 1 would break the repo's existing "every slice green on its own commit" rule (`CLAUDE.md`, "Landing work"; `solid-groove-reviewer.md` §6b), so it would either block its own stack or redden `main` for every parallel task. `test.fixme` is skipped, so PR 1 is honestly green and mergeable, and the marker is what the closing PR removes. The red→green evidence lives in the run log, where it was always more useful than in a merge.
2. **Static screenshots, not video.** Playwright records no audio track, so for a music tool a video is no better than stills; and only GIFs render from an external URL. More importantly, **images cannot be attached to a PR body through the GitHub API at all** — the media-upload endpoint behind drag-and-drop is not in the REST API and `gh` has no command for it. `walkthrough:publish` works around that by pushing PNGs to a `claude/walkthroughs` orphan branch and emitting `raw.githubusercontent.com` links.
3. **The agent self-applies `deploy-preview`.** Accepted, with the risk stated at both call sites: a preview runs against the **live production** backend with production's current rules. The implementer and the workflow are both told never to label a stack that changes `firestore.rules` or `storage.rules`, since a preview cannot prove a rules change and an unreviewed branch must not reach production's access rules.
4. **No GitHub label dispatch.** Dropped per your call — you trigger from claude.ai or iOS. (Worth recording: Routine GitHub triggers only support `pull_request` and `release` events, not `issues`, so a label trigger would have needed an Actions workflow firing a Routine API endpoint anyway.)
5. **`docs/core-flows.md` is the durable QA record**, issues link flows by id, and implementers and reviewers may not edit it or `docs/prd.md`.
6. **The flow spec is reviewed on its own and then frozen.** The reviewer now verifies with `git diff` that neither the register, the PRD, nor a spec's assertions changed across the stack. If the implementer can edit the test, the test proves nothing.

## Core flows

[...]

PR #243, opened 2026-08-06
Map a browsed library sound onto a project asset

afternoon:
## What & why

2 of 5 for #225, builds on #242.

The library's read model and the domain's are different things on purpose: a
`LibraryAsset` is a row of a pack manifest, complete with the genre and character
tags a browser filters on, while a domain `Asset` is the *reference* a project
stores — a fresh `ast_` ID, the pack and version it resolved from, and where the
audio is delivered. Nothing crossed between them, which is the second half of why
a browsed sound cannot reach a track.

`src/library/insertion.ts` is that crossing, and the only place that knows both
shapes:

- `loadSampleCommands` returns the commands that carry the sound and point a
  sampler at it — one transaction, so a drop is one revision and one undo rather
  than an asset stranded in the project when the load is taken back.
- A delivery the project already carries is **reused** rather than duplicated, so
  the same hat on four tracks is one asset and one `AudioBufferCache` entry.
- `librarySampleSchema` is the payload a drag will carry, and it is parsed rather
  than cast: a `dataTransfer` can be written by any page the user drags from.
- An asset with no master audio — a preset — is refused, so it cannot become a
  reference whose buffer never resolves.

Two supporting changes:

- The manifest read model now carries each asset's delivered rights position
  (`license.id`), so a project records real provenance instead of `createAsset`'s
  `internal` default. Relevant to LIB-05, where the rights position is a product
  fact rather than a detail.
- `createAsset`'s `pack` option now asks only for the two fields an asset
  reference actually stores — the pack's `id` and `version`, as a `Pick` of
  `Pack` — rather than a whole `Pack` record. A caller resolving from a pack
  *manifest* holds exactly those and no more: the manifest's asset rows carry no
  rights block. A full `Pack` still satisfies it structurally, so every existing
  caller is unchanged.

Refs #225

## Core flows

Untouched — no user-visible surface in this slice. See #246.

## Walkthrough

No UI change.

## Evidence

```
bun run typecheck   → clean
bun run check       → Checked 502 files. No fixes applied.
bun run test        → Test Files 170 passed (170) · Tests 2278 passed (2278)
```

`insertion.test.ts` drives every case off the committed slice of the *real*
delivered library (`src/library/__fixtures__`), so it parses the wire shape
production parses rather than a hand-written asset that could drift from it.

Invariant worth checking in one line: no existing caller of `createAsset`
changed — `git diff main...HEAD --stat` shows `src/domain/factories.ts` at 7
lines, all inside the `pack` option's type and its comment.

## Acceptance criteria met

#225 has no acceptance checkboxes. This slice supplies the mapping its
"Expected" outcome needs; the surfaces land in PRs 3–5.

## Stack

2. **this PR** — the library → domain asset mapping
3. #244 — let a library row carry its sound on a drag
4. #245 — load a dropped sound onto a sampler

### DIFF
--- a/src/domain/factories.ts
+++ b/src/domain/factories.ts
@@ -419,8 +419,13 @@ export interface CreateAssetOptions {
 	 * The pack this asset resolves from. Required: an asset with no pack is not
 	 * representable (invariant 12), so a factory that defaulted it would be a way
 	 * to create one by accident.
+	 *
+	 * Only the identity and version are asked for, because that is all an asset
+	 * reference stores — and a caller resolving a sound from a *pack manifest*
+	 * holds exactly that, without the rights position or description a full
+	 * `Pack` record carries. A `Pack` satisfies it structurally.
 	 */
-	pack: Pack;
+	pack: Pick<Pack, "id" | "version">;
 	name: string;
 	storageRef: string;
 	kind?: Asset["kind"];
--- a/src/library/audition.test.ts
+++ b/src/library/audition.test.ts
@@ -28,6 +28,7 @@ function asset(overrides: Partial<LibraryAsset> = {}): LibraryAsset {
 		sampleRate: 48000,
 		channelCount: 1,
 		bpm: null,
+		licence: "solid-groove-owned",
 		...overrides,
 	};
 }
--- a/src/library/insertion.test.ts
+++ b/src/library/insertion.test.ts
@@ -1,0 +1,185 @@
+import { describe, expect, it } from "vitest";
+import { executeTransaction } from "../commands";
+import { createFactoryContext } from "../domain/factories";
+import { createSliceFixtureProject } from "../domain/fixtures";
+import { createSeededIdFactory } from "../domain/ids";
+import { fixtureFetcher } from "./__fixtures__/fixtures";
+import {
+	carriedAsset,
+	createLibraryAsset,
+	loadSampleCommands,
+	toLibrarySample,
+} from "./insertion";
+import { LIBRARY_ROOT, packAssets, parsePackManifest } from "./manifest";
+
+/**
+ * Loading a browsed sound onto a track (#225).
+ *
+ * The assets here come from the committed slice of the real delivered library,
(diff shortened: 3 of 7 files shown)


afternoon:
CI red on `browser e2e (firefox)` — **failing on `main` too, and not caused by this stack.**

The failure is `CF-001` at line 95: after clicking "Start playback", `getByRole("button", { name: "Stop playback" })` never appears in Firefox.

That is the known, tracked gap in [`docs/testing.md`](https://github.com/afternoon/solid-groove/blob/main/docs/testing.md#playback-is-asserted-in-chromium-only--a-known-tracked-gap) and #43: in Firefox on a runner, `useProjectAudio.play()` never reaches `setIsPlaying(true)` because `runtime.resume()` neither resolves nor rejects, so the transport button never flips.

Evidence it is not ours:

- `main` at `efdad99e` — [run 31129974187](https://github.com/afternoon/solid-groove/actions/runs/31129974187) — failed with **exactly one** failed job, `browser e2e (firefox)`, same assertion.
- This PR's diff is `src/library/insertion.ts`, `src/library/manifest.ts`, and one type widening in `src/domain/factories.ts`. No component, no audio, no transport. The same job fails identically on the slices above and below it in the stack.

The root cause looks structural rather than flaky: `e2e/smoke.spec.ts:304` and `e2e-emulator/slice.spec.ts:19` both guard this exact assertion with `const canAssertPlayback = browserName === "chromium"`, and `e2e/flows/CF-001.spec.ts` does not. It has been unguarded since #220 landed it, so Firefox has been failing on every branch since.

I have not touched it. `CF-001.spec.ts` is a frozen flow spec and `docs/core-flows.md` is read-only to implementers — adding the guard changes a flow's assertions, which is the product owner's call and its own issue, not something to slip into a bug fix for #225. Flagging it here rather than working around it.

Everything this stack is responsible for is green: `bun run typecheck`, `bun run check`, `bun run test` (2278 passing), and `browser e2e (chromium)`. I will re-check and merge the base branch in when `main` goes green.


---

Issue #224, opened 2026-08-06
No way to change a track's instrument type


afternoon:
The command layer can replace a track's instrument — `instrument.set` with its inverse (`src/commands/definitions/instruments.ts:86`, `changeInstrument` at :130) — but **no component ever dispatches it**:

```
$ grep -rn "changeInstrument" src/ --include=*.tsx | grep -v test
(no matches)
```

`InstrumentPanel` (`src/editor/InstrumentPanel.tsx:36`) switches which panel it renders based on `instrument.kind`, so it displays whichever instrument the track was born with and offers no way to change it.

## Why it matters

Combined with #223 (Add track always creates a synth), a track's instrument is fixed for the life of the track. A producer who adds a track and then wants a sampler on it has to delete it and start over — and with Add track always producing a synth, "start over" doesn't help either.

Changing the instrument on an existing track is also the natural repair path when a user realises a part wants a different sound source, and it keeps the track's mixer settings, devices, sends, and clips rather than discarding them.

## Expected

A track's instrument can be changed from the track's own panel, dispatching the existing `instrument.set` command so the change is one undoable transaction and `instrument_changed` is emitted with the new `instrument_type`.

## Notes

Worth deciding what happens to clips whose content no longer fits the new instrument — a drum machine's pad-keyed clip on a track that becomes a synth, for example. `instrument.set` inverts by restoring the previous instrument, so undo is already correct; the clip question is a product decision, not a command-layer one.

PR #245, opened 2026-08-06
Load a sound dropped from the library onto a sampler

afternoon:
## What & why

4 of 5 for #225, builds on #244. This is the half that takes the drop.

A track's instrument controls are now a drop target for a library sound, and the
editor turns what lands on them into one transaction.

`InstrumentArea` is that target: one region wrapping the kind picker (#224) and
whichever panel the kind gets, named `"Hats instrument"` rather than `"Sampler"`,
because a project with several tracks has several of these and a drop has to land
on a particular one. That is also the name CF-002's spec already uses.

The **area**, rather than the sampler panel alone, for two reasons: the gesture
does not require hitting a small strip, and a `drop` does not travel *down* the
tree — a drop on the picker at the top of the area would otherwise reach no
handler at all. The panels inside keep their own names (`"Sampler"`,
`"Synth voice"`), which is what leaves #224's surface exactly as it was.

It reports what landed rather than dispatching it: the transaction needs the open
project and an ID factory, which belong to the composition root, not to a panel.
`EditorView.loadLibrarySample` is that root's one path — it carries the asset and
points the sampler at it in a single transaction, so a drop is one revision and
one undo, and emits `instrument_changed` once. It loads onto the **selected**
track (#228), not the project's first. Only a sampler is a target at all: a drag
over a synth or drum-machine track shows no drop cursor and dispatches nothing.

**The swap list stays for now.** Retiring it in favour of the loaded sample's
name is #246. Removing it here would take away the only way to change a sample
before its replacement had landed.

Refs #225

## Core flows

Untouched. CF-002 (#61) exercises this surface but stays `test.fixme` until #223
lands too — see #246.

One **non-flow** spec did change, and a reviewer should check it: `e2e/instrument.spec.ts`
(#224's) needed its picker locator made exact —

```ts
const picker = page.getByRole("region", { name: "Instrument", exact: true });
```

Playwright matches an accessible name by *substring*, so once a sibling region is
named `"BD instrument"`, the old `{ name: "Instrument" }` resolved to two
elements. The same applies to its `show(/^Instrument$/)` call. Its assertions are
otherwise untouched: it still asserts `"Sampler"`, `"Synth voice"`, the pads, and
the one-entry-per-switch undo, all unchanged.

## Walkthrough

#246 carries it: this slice adds the drop but leaves the swap list in place, so
the panel a reviewer would see is not the one that ships.

## Evidence

Re-run after rebasing onto `main` at `373e7a9`:

```
bun run typecheck              → clean
bun run check                  → Checked 511 files. No fixes applied.
bun run test                   → Test Files 174 passed (174) · Tests 2339 passed (2339)
bun run test:browser:chromium  → 20 passed, 2 skipped
```

Tests for this slice's own behaviour, at both layers:

- `InstrumentArea.test.tsx` — a dropped sound is reported; a drag carrying
  something else is ignored; a payload that is not a library sound is ignored;
  the area marks itself a copy target only while a sound is over it; a drag it
  cannot take is left to the browser; and a track with **no sampler** is not a
  target at all (no drop cursor, nothing dispatched).
- `editorViewModel.test.ts` — `samplerTrackId` names the track a drop loads onto,
  follows the selected track rather than the project's first, and is null for a
  drum machine, a synth, and no project.
- `EditorView.test.tsx` — a drop loads the sound end to end, emits
  `instrument_changed` exactly once, and **one** undo takes the whole drop back.

## Size — over the ceiling, deliberately

**468 changed lines, against the 400 limit.** Stating it rather than shaving
tests to hide it.

It grew when this stack was rebased onto the moved `main`: satisfying CF-002's
frozen drop-target name *and* #224's landed instrument spec required introducing
`InstrumentArea` (component + tests, ~200 lines) rather than renaming the sampler
panel in place.

I could not find a further split that the project's other rules allow. The
component, its wiring in `TrackEditor`, and `EditorView`'s handler are mutually
dependent: a PR with the component alone leaves `TrackEditor` without a
`loadSample` source, and moving the editor tests into a later PR would ship
untested code and then a tests-only PR — both of which CLAUDE.md forbids more
firmly than the line count. Happy to split it differently if you would rather.

## Acceptance criteria met

#225 has no acceptance checkboxes. This slice delivers the drop half of its
"Expected" outcome: dragging an asset onto a sampler loads it through
`instrument.setSample` as one undoable transaction, emitting `instrument_changed`.

## Stack

2. #243 — the library → domain asset mapping
3. #244 — the drag source
4. **this PR** — the instrument area takes the drop

### DIFF
--- a/e2e/instrument.spec.ts
+++ b/e2e/instrument.spec.ts
@@ -15,17 +15,19 @@ test("changes a track's instrument from its own panel", async ({ page }) => {
 	await page.getByRole("button", { name: "New Project" }).click();
 	await expect(page.getByRole("region", { name: "Step editor" })).toBeVisible();
 
+	// `exact` because the track's instrument controls now sit in a region named
+	// "<track> instrument" (#225), which a substring match would also select.
+	const picker = page.getByRole("region", { name: "Instrument", exact: true });
 	// The workspace scrolls inside a viewport-height app, so each capture
 	// scrolls that container and then returns the page itself to the top —
 	// otherwise the shot is framed on empty page below the editor.
-	const picker = page.getByRole("region", { name: "Instrument" });
 	const show = async (name: string | RegExp): Promise<void> => {
 		await page.getByRole("region", { name }).scrollIntoViewIfNeeded();
 		await page.evaluate(() => window.scrollTo(0, 0));
 	};
 	// The starter track is a sampler, and the panel now says so out loud.
 	await expect(page.getByRole("region", { name: "Sampler" })).toBeVisible();
-	await show("Instrument");
+	await show(/^Instrument$/);
 	await step("Open a project: the track's instrument is a sampler");
 
 	// Sampler -> synth, in one transaction the command layer can name.
--- a/src/editor/EditorView.test.tsx
+++ b/src/editor/EditorView.test.tsx
@@ -8,6 +8,9 @@ import {
 	within,
 } from "@solidjs/testing-library";
 import { afterEach, beforeAll, describe, expect, it, vi } from "vitest";
+import { Analytics } from "../analytics/analytics";
+import { ConsentStore } from "../analytics/consent";
+import { createRecordingTransport } from "../analytics/transport";
 import { installWebAudioGlobals } from "../audio/testAudioContext";
 import {
 	createDrumMachineFixtureProject,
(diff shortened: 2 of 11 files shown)


afternoon:
CI red on `browser e2e (firefox)` — pre-existing on `main`, not this PR. 1 failed job of 9, `e2e/flows/CF-001.spec.ts:95`; 17 passed, 2 skipped. `main` at `efdad99e` fails on exactly the same single job.

Tracked in #249 with the full evidence and a suggested fix. Not fixing it from here: it changes a frozen flow spec's assertions.

Relevant to this slice specifically, since it is the one that adds the drop target: `browser e2e (chromium)` and `browser e2e (webkit)` both pass, alongside typecheck, lint and the unit suite (2286 passing). The Firefox job fails at CF-001 step 7 — a flow this PR does not touch — before reaching anything here.


---

PR #246, opened 2026-08-06
Retire the sampler's swap list, and insert from the keyboard

### LINKED ISSUE #225: A library sound cannot be loaded onto a sampler — drag from the browser to the instrument

afternoon:
## What & why

5 of 5 for #225, builds on #245. This one completes the task.

The swap list could only ever offer sounds the project already carried — for a
new project, the single one it started with — which is exactly why the issue
calls the library "200 sounds a user can hear and cannot use". Now that a sound
can be dragged onto the sampler, the panel says what it is holding and how to
change it, and the list goes. `replacementOptions` goes with it: it was the last
consumer of the project's own sample list as a menu.

Dragging is a pointer gesture, so it cannot be the only way in (PRD 9.3). The
library row's "Insert" button was already built and simply never handed a
handler; wiring it to the same `EditorView.loadLibrarySample` the drop uses means
the keyboard path produces the same transaction, the same undo, and the same
event, rather than a second implementation that can drift.

Closes #225

## Core flows

Untouched. #225 links no core flow of its own. CF-002 (#61) leans on this
affordance in its `addSamplerTrack` helper and stays `test.fixme`, because its
other dependency — creating a sampler track, #223 — has not landed. Nothing in
`docs/core-flows.md` or in any flow spec changed in this stack; `bun run
verify:core-flows` passes and reports CF-002 and CF-003 still parked.

(One **non-flow** spec changed, in #245: `e2e/instrument.spec.ts`'s picker
locator had to become `exact`. Explained there.)

Two things CF-002 will need, worth flagging on #61 rather than working around
here:

- Its selectors for this half now exist and match what shipped: the row is a
  draggable `listitem`; the drop target is a `region` whose accessible name is
  the track's name followed by the word `instrument` (so `"Hats instrument"`);
  and it contains the loaded sound's name.
- **The delivered library is not present in a browser E2E run.** `public/samples`
  is gitignored and produced by `bun run library:build`, which CI does not run, so
  the library panel in `bun run test:browser` shows its empty/error state. That
  predates this stack, but CF-002 cannot pass until it is addressed.

## Walkthrough

Captured in Chromium against a locally built delivered library
(`bun run library:build`), and published with `bun run walkthrough:publish`.

**Not** captured from a core-flow spec, because the flow that covers this
(CF-002) is blocked on #223 and cannot run. It was captured by driving the same
steps from the same entrypoint the flow uses, so it shows real shipped behaviour;
the flow-derived walkthrough arrives with #61.

### A producer loads a library sound onto a sampler

**1. New project, on the starter kick**

issue-225 step 1

**2. Search for a clap**

issue-225 step 2

**3. Drop it on the sampler**

issue-225 step 3

**4. One undo takes it back**

issue-225 step 4

**5. Insert from the keyboard**

issue-225 step 5

These were captured before the rebase onto `main` at `373e7a9`. The drop target's
accessible name is unchanged by that rebase, and the flow they show still passes
— but they do not show the kind picker (#224) that now sits above the panel, so
treat them as the interaction, not the current pixel layout.

## Evidence

Re-run after the rebase:

```
bun run typecheck              → clean
bun run check                  → Checked 511 files. No fixes applied.
bun run test                   → Test Files 174 passed (174) · Tests 2336 passed (2336)
bun run verify:core-flows      → 3 flow(s), each with exactly one spec (CF-002/CF-003 parked)
bun run test:browser:chromium  → 20 passed, 2 skipped
```

Chromium-only for the browser suite — this environment cannot reach
`cdn.playwright.dev` to install Firefox or WebKit, so that run is **not** the PRD
section 10 gating evidence; CI's full matrix on push is.

Diff: 290 changed lines.

## Acceptance criteria met

#225 has no acceptance checkboxes; against its "Expected" and "Notes":

- ✅ Dragging an asset from the library onto a sampler instrument loads it,
  dispatching `instrument.setSample` as one undoable transaction and emitting
  `instrument_changed`. (`asset.add` rides in the same transaction, since a
  browsed sound is not yet one of the project's assets.)
- ✅ The swap list is dropped in favour of a label showing the loaded sample's
  name, as the product owner asked.
- ✅ Drag-and-drop has a keyboard-reachable equivalent: the browser's existing
  `onInsert`, loading onto the track being edited, through the same code path.

Not in scope, and deliberately left: loading onto a **drum pad**
(`drum.setPadAsset` has the same gap, but the issue names the sampler). Choosing
*which* track a keyboard insert reaches is no longer a gap — after rebasing onto
#228's selected-track work, both the drop and the insert follow the selected
track.

## Stack

2. #243 — the library → domain asset mapping
3. #244 — the drag source
4. #245 — the instrument area takes the drop
5. **this PR** — retire the swap list, insert from the keyboard (`Closes #225`)

### DIFF
--- a/src/editor/EditorView.test.tsx
+++ b/src/editor/EditorView.test.tsx
@@ -18,8 +18,10 @@ import {
 	createSliceFixtureProject,
 } from "../domain/fixtures";
 import { fakePreviewEngine } from "../library/__fixtures__/fakePreviewEngine";
+import { fixtureFetcher } from "../library/__fixtures__/fixtures";
 import { LIBRARY_SAMPLE_MIME } from "../library/assetDrag";
 import type { PreviewEngine } from "../library/audition";
+import { LibraryClient } from "../library/libraryClient";
 import type { InMemoryProjectRepository } from "../persistence/inMemoryProjectRepository";
 import { detectPlatform, shortcutLabel } from "../shortcuts";
 import { memoryStorage } from "../testing/storage";
@@ -104,6 +106,7 @@ function renderEditor(
 	projectId: string,
 	options: {
 		createAuditionEngine?: () => PreviewEngine;
+		libraryClient?: LibraryClient;
 		analytics?: Analytics;
 	} = {},
 ) {
@@ -116,6 +119,7 @@ function renderEditor(
 					<EditorView
 						projectId={projectId}
 						createAuditionEngine={options.createAuditionEngine}
+						libraryClient={options.libraryClient}
 						analytics={options.analytics}
 					/>
 				)}
@@ -265,12 +269,11 @@ describe("EditorView", () => {
 		expect(
 			screen.getByRole("button", { name: "Audition" }),
 		).toBeInTheDocument();
-		// Scoped to the name paragraph: the swap list this PR leaves in place
-		// carries the same name in an <option>.
+		// It says what it is holding rather than offering a list of the project's
+		// own samples to swap between (#225).
 		const panel = screen.getByRole("region", { name: "BD instrument" });
-		expect(
-			within(panel).getByText("909 Bass Drum", { selector: "p" }),
-		).toBeInTheDocument();
+		expect(within(panel).getByText("909 Bass Drum")).toBeInTheDocument();
+		expect(within(panel).queryByRole("combobox")).not.toBeInTheDocument();
 	});
 
 	it("loads a sound dropped from the library onto the sampler, undoably", async () => {
@@ -290,9 +293,7 @@ describe("EditorView", () => {
 		fireEvent.drop(panel, { dataTransfer: transferCarrying(DROPPED_HAT) });
 
 		// The sampler names the dropped sound, and the project now carries it.
-		expect(
-			await within(panel).findByText(DROPPED_HAT.name, { selector: "p" }),
-		).toBeInTheDocument();
+		expect(await screen.findByText(DROPPED_HAT.name)).toBeInTheDocument();
 		const changed = transport.named("instrument_changed");
 		expect(changed).toHaveLength(1);
 		expect(changed[0].params.instrument_type).toBe("sampler");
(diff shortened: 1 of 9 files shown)


afternoon:
CI red on `browser e2e (firefox)` — same pre-existing `main` failure as the rest of the stack; diagnosed in full at #243 (comment) and tracked as #43. `CF-001.spec.ts:95` asserts the transport flips to "Stop playback", which never happens in Firefox on a runner, and `CF-001` is the one spec that does not guard that assertion with `browserName === "chromium"`. `main` at `efdad99e` fails on exactly the same single job.

Worth recording here specifically, because this is the stack's UI PR: on the same run, **`browser e2e (chromium)` and `browser e2e (webkit)` both passed**, along with typecheck and the unit suite — 1 failed job out of 8. So renaming the sampler panel's accessible name from `"Sampler"` to `"<track> instrument"`, removing the swap list, and wiring the drop broke no existing browser spec in either of the two browsers that can run these assertions.

Not fixing it from here: `CF-001.spec.ts` is a frozen flow spec, and adding the Chromium guard changes a flow's assertions. I will merge `main` in and let CI re-run once it is green.


---

PR #248, opened 2026-08-06
Declare the instrument kinds once, for both surfaces

### LINKED ISSUE #247: A created drum machine and a switched-to drum machine get different pads
#232 (for #223) and #234 (for #224) were written in parallel and merged two minutes apart. Both needed the same facts — which instrument kinds exist, what each is called, what `instrument_type` each reports, and what instrument each mints — and each landed its own copy.

`src/editor/trackCreation.ts` (creating a track):

```
const DEFAULT_PAD_NAMES = ["BD", "SD", "HH", "CP"] as const;
```

`src/instrument/instrumentKinds.ts` (changing an existing track's instrument):

```
export const DEFAULT_DRUM_PAD_NAMES: readonly string[] = ["Kick", "Snare", "Hat", "Clap"];
```

## Why it matters

The divergence is visible to a user, not just to a reader. **Add a drum machine track and its pads are `BD / SD / HH / CP`; switch an existing track to a drum machine and they are `Kick / Snare / Hat / Clap`** — the same product decision answered two ways, depending on which door you came through.

The rest of the duplication has not diverged yet but has nothing holding it together: two kind tables (`NEW_TRACK_KINDS`, `INSTRUMENT_KIND_CHOICES`) already disagree on the order the three kinds are offered in, two `instrumentTypeKey` functions map a kind to the same three analytics values with different signatures, and two builders construct the same instruments.

## Expected

One declaration of the instrument kinds, their labels, their `instrument_type`, and the instrument each mints, read by both surfaces — so a drum machine is the same drum machine however a user arrives at one, and a fourth kind is added in one place.

afternoon:
## What & why

#232 (for #223) and #234 (for #224) were written in parallel and merged two minutes apart. Both needed the same facts — which instrument kinds exist, what each is called, what `instrument_type` each reports, and what instrument each mints — and each landed its own copy. The copies had already diverged where a user could see it: a track *created* as a drum machine opened with `BD / SD / HH / CP`, one *switched* to a drum machine with `Kick / Snare / Hat / Clap`.

`src/instrument/instrumentKinds.ts` now declares each kind once — label, `instrument_type`, and the instrument it mints — and both surfaces read it. `src/editor/trackCreation.ts` keeps what is only track creation's: the button's accessible name (now derived from the shared label rather than written out beside it) and the clip and placement a new track opens with.

**Its exported surface is unchanged** — `NEW_TRACK_KINDS`, `NewTrackKind`, `NewTrackKindSpec`, `newTrackKindSpec`, `createNewTrack`, and `instrumentTypeKey` all still resolve, the last as a re-export. That is deliberate: the mixer half of #223 is not landed yet, and this must not break the branch it is on.

Closes #247

### Two consequences, stated rather than left to be found

- **A track switched to a drum machine now opens with `BD / SD / HH / CP`**, adopting what creating one already gave. Picking the other set would have changed a landed feature's behaviour, which is beyond a dedupe — but `Kick / Snare / Hat / Clap` is arguably the friendlier label set for a tool aiming at "accessible and intuitive", and it is now a one-line change in `DEFAULT_PAD_NAMES` if that call is wrong.
- **The picker now offers the shared order** — sampler, drum machine, synth — rather than its own sampler, synth, drum machine. One order for the same three choices.

`e2e/instrument.spec.ts` changes its two `Audition Kick` assertions to `Audition BD` as a direct consequence of the first bullet. That is the only assertion change in this PR, and it follows the behaviour rather than accommodating it.

## Core flows

None — #247 is a bug report and links no core flow ID. `bun run verify:core-flows` passes and reports the three registered flows unchanged; no flow spec was touched.

## Walkthrough

Recaptured from the passing `e2e/instrument.spec.ts` after this change, in Chromium. The visible difference is step 3: the pads now read `BD / SD / HH / CP`, and the picker lists drum machine second.

### issue-224 — Change a track's instrument

**1. Open a project: the track's instrument is a sampler**

issue-224 step 1

**2. Pick Synth: the synth panel replaces the sampler's**

issue-224 step 2

**3. Pick Drum machine: it arrives with pads to load**

issue-224 step 3

**4. Undo returns the drum machine, one switch at a time**

issue-224 step 4

<sub>Captured by `bun run walkthrough:capture` and published to the `claude/walkthroughs` branch by `bun run walkthrough:publish`.</sub>

> **`walkthrough:publish` emits URLs that cannot be posted from an agent session.** A `raw.githubusercontent.com` URL is rewritten to backtick-wrapped text on the way into a PR body or comment, so the image renders as an empty code span — that is why #235 merged with four broken images, and why the stray comment below is unreadable. The same file served from the `github.com` `/raw/refs/heads/...` path is not filtered, which is what this section uses. `scripts/walkthrough/publish.mjs` should emit that form instead; happy to raise it as its own issue.

## Evidence

```
$ bun run check:ci
Checked 490 files in 624ms. No fixes applied.

$ bun run test
Test Files  170 passed (170)
     Tests  2278 passed (2278)

$ bun run test:browser:chromium -- e2e/instrument.spec.ts
1 passed (15.1s)
  ✓ e2e/instrument.spec.ts:8:1 › changes a track's instrument from its own panel

$ bun run verify:core-flows
check-core-flows — 3 flow(s), each with exactly one spec.
```

The regression test that makes this stick, in `instrumentKinds.test.ts`, builds the instrument **both ways** for every kind and asserts they are equal:

```ts
const created = createNewTrack(context(), { kind: spec.kind, ... }).track.instrument;
const switched = createInstrumentOfKind(context(), spec.kind);
expect(created, spec.kind).toEqual(switched);
```

It fails on `main` as it stands (`Kick` ≠ `BD`), which is the bug in #247. `src/editor/trackCreation.test.ts` is untouched and its 6 tests pass unchanged — including the WCAG 2.5.3 check that each button's visible label is inside its accessible name, which the derived `actionLabel` still satisfies.

Firefox and WebKit are CI's to gate: this environment cannot reach `cdn.playwright.dev`, so the Chromium run is a pre-flight, not PRD section 10 evidence. Note also that `main` is currently red on `browser e2e (firefox)` at `e2e/flows/CF-001.spec.ts:95` — the documented `AudioContext.resume()` gap, unrelated to this change.

## Acceptance criteria met

- ✅ One declaration of the kinds, their labels, their `instrument_type`, and the instrument each mints, read by both surfaces.
- ✅ A drum machine is the same drum machine whichever door the user came through, with a test that fails if the two drift apart again.
- ✅ A fourth kind would now be added in one place.
- ✅ No behaviour change beyond the two consequences named above; `trackCreation.ts`'s exported surface is unchanged so the unlanded #223 mixer PR still compiles.

### DIFF
--- a/e2e/instrument.spec.ts
+++ b/e2e/instrument.spec.ts
@@ -38,13 +38,12 @@ test("changes a track's instrument from its own panel", async ({ page }) => {
 	).toBeEnabled();
 	await step("Pick Synth: the synth panel replaces the sampler's");
 
-	// ...and on to the drum machine, which arrives with pads to load.
+	// ...and on to the drum machine, which arrives with the same pads a track
+	// created as a drum machine gets (BD/SD/HH/CP, from the shared kind table).
 	await picker.getByText("Drum machine", { exact: true }).click();
-	await expect(
-		page.getByRole("button", { name: "Audition Kick" }),
-	).toBeVisible();
+	await expect(page.getByRole("button", { name: "Audition BD" })).toBeVisible();
 	await show(/^Drum machine/);
-	await step("Pick Drum machine: it arrives with pads ready for sounds");
+	await step("Pick Drum machine: it arrives with pads to load");
 
 	// The route back off a drum machine — the direction that had no UI at all,
 	// since the drum machine's own panel is mounted elsewhere.
@@ -53,9 +52,7 @@ test("changes a track's instrument from its own panel", async ({ page }) => {
 
 	// Each switch is one history entry, so undo walks back a step at a time.
 	await page.getByRole("button", { name: /^Undo/ }).click();
-	await expect(
-		page.getByRole("button", { name: "Audition Kick" }),
-	).toBeVisible();
+	await expect(page.getByRole("button", { name: "Audition BD" })).toBeVisible();
 	await show(/^Drum machine/);
 	await step("Undo returns the drum machine, one switch at a time");
 });
--- a/src/editor/trackCreation.ts
+++ b/src/editor/trackCreation.ts
@@ -1,77 +1,55 @@
-import type { InstrumentTypeKey } from "../analytics/catalog";
-import type { Clip, Instrument, Placement, Track } from "../domain/entities";
+import type { Clip, Placement, Track } from "../domain/entities";
 import {
-	createDrumMachineInstrument,
-	createDrumPad,
 	createNoteClip,
 	createPlacement,
-	createSamplerInstrument,
-	createSynthInstrument,
 	createTrack,
 	type DomainFactoryContext,
 } from "../domain/factories";
 import { TICKS_PER_BAR } from "../domain/time";
+import {
+	createInstrumentOfKind,
+	INSTRUMENT_KINDS,
+	type InstrumentKind,
+	type InstrumentKindSpec,
+} from "../instrument/instrumentKinds";
 
 /**
  * What creating a track produces, and the kinds it can be created as (#223).
  *
  * Track creation used to mint a synth and nothing else, so a sampler or a drum
- * machine could only ever arrive with the starter project or a fixture. The
- * kinds live in one table here — framework-free and testable without rendering
- * the mixer — so the affordance that offers them, the instrument each mints,
- * and the analytics value each reports cannot drift apart.
+ * machine could only ever arrive with the starter project or a fixture. Which
+ * kinds exist, what each is called, what instrument each mints, and what each
+ * reports to analytics are `src/instrument/instrumentKinds.ts`'s to say — the
+ * instrument panel's kind picker chooses from the same set, and the two had
+ * begun to disagree. What is track creation's alone stays here: the button's
+ * accessible name, and the clip and placement a new track opens with.
  *
  * Nothing here mutates a project: it builds the entities a `track.add` command
  * carries, so the mixer's dispatch stays the one mutation path (PRD 9.6).
  */
 
 /** The instrument kinds a new track can be created with. */
-export type NewTrackKind = "sampler" | "drumMachine" | "synth";
+export type NewTrackKind = InstrumentKind;
 
-export interface NewTrackKindSpec {
-	readonly kind: NewTrackKind;
-	/** The button's visible text. */
-	readonly label: string;
+export interface NewTrackKindSpec extends InstrumentKindSpec {
 	/**
-	 * The button's accessible name. It contains {@link label} so the visible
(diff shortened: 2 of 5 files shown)


afternoon:
## Walkthrough

The PR body's copy of this did not render — image markdown is corrupted in transit when a body is written through the API tooling available here (the same thing happened to #235, which merged with four broken images). Reposting as a comment to find out whether comments survive it.

Recaptured from the passing `e2e/instrument.spec.ts` after this change, in Chromium.

### issue-224 — Change a track's instrument

**1. Open a project: the track's instrument is a sampler**

issue-224 step 1``

**2. Pick Synth: the synth panel replaces the sampler's**

issue-224 step 2``

**3. Pick Drum machine: it arrives with pads ready for sounds** — note the pads now read `BD / SD / HH / CP`, matching what creating a drum machine track already gave.

issue-224 step 3``

**4. Undo returns the drum machine, one switch at a time**

issue-224 step 4``

---

afternoon:
**CI: `browser e2e (firefox)` is red, inherited from the base branch — not this change.** Every other job in [run 31129680018](https://github.com/afternoon/solid-groove/actions/runs/31129680018) passed (typecheck/`check:ci`/unit+component, emulator suite, emulator e2e in both browsers, browser e2e chromium, production build; webkit still running).

The single failure is `e2e/flows/CF-001.spec.ts:95` waiting for `Stop playback` — 17 passed, 2 skipped, 1 failed, across all three retries. That is the documented Firefox `AudioContext.resume()` gap, not a shortcut- or instrument-related failure: see `docs/testing.md`, "Playback is asserted in Chromium only — a known tracked gap".

It reproduces on `main` at exactly the same job, spec, and line — [run 31128968506](https://github.com/afternoon/solid-groove/actions/runs/31128968506) at `38c81ae`, before any of this stack existed. So it predates this PR and is not something to fix here; CF-001 is a frozen flow spec and its assertion is correct — the gap is in the product, not the test.

Two things a reviewer should know rather than infer:

1. **This run is against `e3ef430`, which is no longer this PR's head.** The branch was rewritten and force-pushed twice since (current head `8731a4d`), so this run validated the earlier single-module version, *not* the deduplication in the current diff. No run has appeared for `8731a4d` — the PR reports zero check runs — so **this PR has not been CI-validated as it stands**, and the green jobs above should not be read as covering it. The local evidence in the description is what I have for the current head.
2. `e2e/instrument.spec.ts` did not exist at `e3ef430`, so the browser coverage for this feature is not in this run either. It passes locally in Chromium (`1 passed (15.1s)`).

Happy to push an empty commit to force a run against the real head if that is the convention here — say the word rather than my adding noise to the history unasked.

---

PR #261, opened 2026-08-18
Fix #259 (1/1): Let a preview channel read the library bucket

### LINKED ISSUE #259: Library doesn't load on PR preview channels - CORS error

afternoon:
## What &amp; why

The library browser never loads on a PR preview channel, reported while walking the #225 stack with #246 at the top. **It is not a URL-encoding bug**: the Firebase Storage download endpoint takes the object path as a single encoded path segment, so `.../o/library%2Fpacks%2Findex.json?alt=media` is the correct form — that URL returns `200` and valid JSON, while the unencoded `.../o/library/packs/index.json` returns `400`.

The real cause is the bucket's CORS policy. `storage.cors.json` named four origins — the project's `web.app` and `firebaseapp.com` domains plus `localhost`/`127.0.0.1`. A Hosting preview channel is none of them: it is served from `https://PROJECT--pr-N-HASH.web.app` (for this one, `https://groove-35c07--pr-246-4wgdvx7q.web.app`), and the trailing hash is minted at deploy time. Cloud Storage matches a CORS origin as an **exact string** with no subdomain wildcard, so no list can name a channel ahead of time. The bucket answers a preview's `packs/index.json` request with a `200` and **no** `Access-Control-Allow-Origin` header, the browser discards the response, and `fetch` rejects — which `libraryClient.ts` classifies as `network` and the browser surfaces as a load failure.

Reproduced against the live bucket — same object, same URL, only the `Origin` header differs:

```
$ curl -sD- -o/dev/null -H 'Origin: https://groove-35c07.web.app' \
    'https://firebasestorage.googleapis.com/v0/b/groove-35c07.firebasestorage.app/o/library%2Fpacks%2Findex.json?alt=media'
HTTP/2 200
access-control-allow-origin: https://groove-35c07.web.app

$ curl -sD- -o/dev/null -H 'Origin: https://groove-35c07--pr-246-4wgdvx7q.web.app' ...
HTTP/2 200
(no access-control-allow-origin header at all)
```

The fix admits any origin for `GET`/`HEAD`. That grants nothing new: `storage.rules` makes `library/**` world-readable and denies every other path, no write method is offered, and these requests carry no credentials — the same bytes already answer an unauthenticated `curl`. Removing the placeholder list also makes the checked-in file directly usable with `gcloud storage buckets update --cors-file=`, once the `_comment` key is stripped.

`corsAllowsOrigin()` writes down the matching rule Cloud Storage actually applies, so the regression test states the symptom — a preview channel origin is admitted — instead of asserting on the policy's text, and pins the no-subdomain-pattern behaviour that caused this.

⚠️ **Merging this does not fix any live preview.** No workflow applies the policy; the bucket has to be re-configured once with `bun run library:upload -- --configure-bucket` (or the equivalent `gcloud storage buckets update gs://groove-35c07.firebasestorage.app --cors-file=` against a copy of the policy with `_comment` removed). Both need the deploy credential, so this is a maintainer step.

PRD LIB-05; sample-library section 12. 1 of 1 — this does not build on the #225 stack and is independent of it.

Closes #259

## Core flows

Untouched. This is a bug fix in delivery configuration, not a feature slice: no flow is specced, completed, or edited, and `docs/core-flows.md`, `docs/prd.md`, and every spec are unchanged.

## Walkthrough

No UI change. Nothing in `src/` changes; the diff is the bucket policy, the publish script, its tests, and two docs. The user-visible effect — the library browser loading on a preview — cannot be captured here because it depends on the policy being applied to the bucket, which this merge does not do.

## Evidence

Regression test red before the fix, against the old policy (`git stash push storage.cors.json`):

```
FAIL scripts/starter-library/upload.test.mjs > CORS configuration >
     admits a Hosting preview channel, whose subdomain cannot be enumerated
AssertionError: expected false to be true
Test Files  1 failed (1)
     Tests  1 failed | 5 passed | 16 skipped (22)
```

Green after:

```
$ bunx vitest run scripts/starter-library/upload.test.mjs
✓ scripts/starter-library/upload.test.mjs (22 tests) 18ms
Test Files  1 passed (1)      Tests  22 passed (22)

$ bun run test
Test Files  170 passed (170)  Tests  2301 passed (2301)   Duration 95.82s

$ bun run check          # runs tsc --noEmit && biome check
Checked 490 files in 1108ms.
```

CI is green on both runs of `492d785`: unit + component, firebase emulator, browser e2e across chromium/firefox/webkit, browser e2e against the emulator, and the production build + no-secrets check. The one red job in the first run was a nondeterministic failure in `src/audio/devices/shaping.test.ts`, unrelated to this diff and diagnosed in the comment below — the same commit passed that job concurrently, and passed again on re-run.

Emulator and browser suites cannot exercise bucket CORS at all — the Storage emulator answers `setCorsConfiguration` with "Not Implemented" and serves permissive CORS regardless (docs/sample-library.md), and the browser suites run against the in-memory backend with no bucket. The live-bucket `curl` transcript above is the evidence for the deployed behaviour.

Diff is 89 changed lines across 5 files, none of them in `src/`.

## Acceptance criteria met

Issue #259 carries no checkboxes. What it reported — the library failing to load on a PR preview channel with a CORS error — is diagnosed, has a test that fails without the fix, and is corrected at its cause. The one thing left, which this PR cannot do, is applying the policy to the bucket; the ⚠️ note above and `docs/testing.md` both say so rather than implying the merge is sufficient.


afternoon:
CI red on `typecheck, check:ci, unit + component tests` — not this PR, and not a flake I am waving through. The same commit `492d785` **passed the identical job** in the concurrent run: [32118086912](https://github.com/afternoon/solid-groove/actions/runs/32118086912/job/95652057387) green, [32117992198](https://github.com/afternoon/solid-groove/actions/runs/32117992198/job/95651762690) red, same SHA, same runner image. This PR also changes nothing under `src/` — the diff is `storage.cors.json`, `scripts/starter-library/upload.mjs`, its test, and two docs — so there is no path by which it reaches an audio DSP test.

The failure is `src/audio/devices/shaping.test.ts:161`, `overdrive and saturator (FX-01, FX-02) > adds harmonics as drive climbs`:

```
AssertionError: expected 0.06561431610569629 to be greater than 1.4142872250844178
```

Worth recording precisely, because the numbers say this is a degenerate render rather than a marginal threshold. `brightness = hfEnergy/rms`, and `hfEnergy` is the RMS of the first difference, so for a pure sine it is exactly `2·sin(π·f/fs)`:

| | brightness |
|---|---|
| clean 220 Hz sine @ 44.1 kHz (expected) | 0.031 |
| `pushed` (drive 0.5), observed | 0.066 |
| `clean` (drive 0), **observed** | **1.414** |

Inverting the identity, `1.4142872 = 2·sin(π·0.250017)` — the drive-0 buffer came back as content at almost exactly **a quarter of the sample rate**. `pushed` at 0.066 is a perfectly sane distorted 220 Hz sine, so the run's `destroyed`/`pushed` renders were fine and only the drive-0 `Tone.Offline` render was garbage. That points at the offline render harness (a wrong-rate or otherwise degenerate context for one render), not at the overdrive curve or the assertion's tolerance.

Not fixing it from here: it is a pre-existing defect in a test this PR does not touch, in a different area of the codebase, and folding an audio-harness fix into a bucket-CORS change would give a reviewer two unrelated things to hold at once. It wants its own issue and `/fix`. I have re-run the failed jobs so this PR's status reflects the code rather than the roll of the dice.

---

PR #262, opened 2026-08-18
Fix #225 (1/1): Claim the drag on arrival, so the area is a drop target at all

afternoon:
## What & why

Reported: dragging a sound out of the library **sidebar** picks it up correctly,
but nothing in the app shows a drop affordance — not the instrument area, the
mixer, or the timeline — and the release does nothing.

**Cause: `InstrumentArea` cancelled `dragover` but never `dragenter`.**

The HTML drag-and-drop processing model chooses the element a drag is pointing
at when `dragenter` fires. Cancel that event and the element becomes the drag's
current target; leave it uncancelled and the current target falls back to **the
body element** — after which `dragover` is fired at the body, and this area's
`dragover` handler never runs at all. No affordance goes up, and there is
nothing droppable anywhere on the page. That is the reported symptom exactly,
including "no drop styling on *any* element".

Chromium deviates here: it keeps firing `dragover` at the element under the
pointer whether or not `dragenter` was cancelled, so cancelling only `dragover`
looks correct there. That is why this passed every check in this repo —
including a real-browser drag I drove through Chromium — and failed for the
reporter. MDN states the requirement plainly: to specify a drop target you
cancel **both** `dragenter` and `dragover`.

Both events now run through one `claim()`, so a drag the track cannot take
(a synth, a drum machine, a payload that is not a library sound) is still left
entirely alone — no drop cursor, and the browser's own handling undisturbed.

**Second, smaller defect, found while reproducing: the affordance flickered
off.** `dragleave` bubbles, so every step from one panel inside the area to the
next reported a leave the drag had never made. Measured in Chromium, sweeping
into the area toggled the outline `[true, false, true, false]`; it is now
`[true]` — up when the drag arrives, and still up at the moment of release.
`relatedTarget` distinguishes leaving the area from moving about inside it.

Refs #225

### Not in this PR

- **The mixer strip and the arrangement timeline row are not drop targets.**
  The reporter expects a drop on either to load the sample; today
  `InstrumentArea` is the only `onDragOver` in the app, so those surfaces were
  never built to accept a drag. That is a feature on top of #225, not a defect
  in it, and it wants its own issue.
- **The pack browser's rows offer a drag nothing can receive** — its modal
  covers every drop target, so a drag started there can only be abandoned. Real,
  reproduced, and unrelated to this report; removing an affordance is the
  product owner's call, so it is not landed here. (It was in the first revision
  of this branch and has been reverted.)

## Core flows

Untouched. CF-002 covers this gesture and stays `test.fixme` — it also needs
"Add sampler track" (#223), which is not in this stack. No flow spec, no
`docs/core-flows.md` and no `docs/prd.md` line was edited. `bun run
verify:core-flows` passes: 3 flows, each with exactly one spec.

## Walkthrough

No walkthrough: this PR closes no issue and delivers no new flow. The visible
change is that a track's instrument area now shows its drop outline when a
library sound is dragged over it, and holds it until the drag leaves.

## Evidence

Both defects have a test that fails first.

```
FAIL src/editor/InstrumentArea.test.tsx > InstrumentArea > becomes a drop target the moment the drag arrives
  AssertionError: expected true to be false
  (fireEvent returns true — the dragenter was never cancelled, so the area was
   never claimed as the drag's target)

FAIL src/editor/InstrumentArea.test.tsx > InstrumentArea > keeps the affordance while the drag crosses its own contents
  AssertionError: expected false to be true
```

Real-browser drag in Chromium, driving the app from the library sidebar to the
sampler panel — before the `dragleave` fix, then after both fixes:

```
before:  TRANSITIONS: [true,false,true,false]
after:   TRANSITIONS: [true]   highlighted=true loaded=true
```

Green after:

```
$ bun run typecheck
(no output)

$ bun run check
Checked 498 files in 1205ms. No fixes applied.

$ bun run test
Test Files  174 passed (174)
     Tests  2340 passed (2340)
```

**A caveat that matters for this particular fix:** this environment cannot
reach `cdn.playwright.dev`, so only Chromium is installed — and Chromium is
precisely the browser that masks this bug. The `dragenter` diagnosis rests on
the specification and on the reported symptom matching its predicted failure
mode ("nothing droppable anywhere"), not on a local reproduction in a browser
that fails. CI runs Firefox and WebKit on this branch, but no browser suite
exercises drag-and-drop yet — CF-002 is still parked — so a green matrix is not
evidence for this fix either. The preview channel, in the reporter's own
browser, is the check that counts.

## Acceptance criteria met

None closed — this is a fix on top of the open stack rather than a slice of
#225, so it carries `Refs` and not `Closes`. It removes the reason a drop from
the library sidebar never registered.

### DIFF
--- a/src/editor/InstrumentArea.test.tsx
+++ b/src/editor/InstrumentArea.test.tsx
@@ -115,6 +115,37 @@ describe("InstrumentArea", () => {
 		expect(loadSample).not.toHaveBeenCalled();
 	});
 
+	it("becomes a drop target the moment the drag arrives", () => {
+		renderArea();
+		const transfer = transferCarrying(JSON.stringify(DROPPED));
+		// A browser decides what a drag may be dropped on from `dragenter` as
+		// well as `dragover` (the HTML drag-and-drop processing model): an
+		// uncancelled `dragenter` sends the pointer's target back to the body,
+		// and this element's `dragover` is then never consulted at all. Chromium
+		// is lenient about that first step and accepts the drop anyway; a browser
+		// that follows the model shows no affordance and takes no drop. So the
+		// arrival has to be claimed too. `fireEvent` returns false when a handler
+		// cancelled the event, which is what claiming it means.
+		const allowed = fireEvent.dragEnter(area(), {
+			dataTransfer: transfer,
+			cancelable: true,
+		});
+		expect(allowed).toBe(false);
+		expect(transfer.dropEffect).toBe("copy");
+		expect(isTarget()).toBe(true);
+	});
+
+	it("leaves an arriving drag it cannot take alone", () => {
+		renderArea(SYNTH);
+		const transfer = transferCarrying(JSON.stringify(DROPPED));
+		const allowed = fireEvent.dragEnter(area(), {
+			dataTransfer: transfer,
+			cancelable: true,
+		});
+		expect(allowed).toBe(true);
+		expect(isTarget()).toBe(false);
+	});
+
 	it("marks itself a copy target only while a sound is over it", () => {
 		renderArea();
 		const transfer = transferCarrying(JSON.stringify(DROPPED));
--- a/src/editor/InstrumentArea.test.tsx
+++ b/src/editor/InstrumentArea.test.tsx
@@ -51,9 +51,30 @@ function renderArea(instrument: Instrument | null = SAMPLER) {
 			loadSample={loadSample}
 		>
 			<p>panel goes here</p>
+			<p>and another panel</p>
 		</InstrumentArea>
 	));
 	return { loadSample };
+}
+
+/**
+ * A `dragleave` that says where the pointer went next.
+ *
+ * jsdom implements no `DragEvent`, so `fireEvent.dragLeave` quietly drops
+ * `relatedTarget` — the one field that distinguishes stepping between the
+ * panels inside the area from leaving it. A `MouseEvent` of the same name
+ * carries it, and in a browser `DragEvent` inherits the field from `MouseEvent`
+ * anyway, so this dispatches exactly what the handler reads.
+ */
+function leaveTowards(from: HTMLElement, to: EventTarget | null): void {
+	fireEvent(
+		from,
+		new MouseEvent("dragleave", {
+			bubbles: true,
+			cancelable: true,
+			relatedTarget: to,
+		}),
+	);
 }
 
 /** The area, found the way a drag target is: by its accessible name. */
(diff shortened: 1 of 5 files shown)


afternoon:
CI red on `typecheck, check:ci, unit + component tests`, and it is not this branch's change — but it is not a flake to wave through either, so here is the diagnosis.

**The failure**

```
FAIL src/audio/devices/shaping.test.ts > overdrive and saturator (FX-01, FX-02)
     > adds harmonics as drive climbs, from coloration to destruction
AssertionError: expected 0.06561431610569629 to be greater than 0.12500782688579118
  ❯ src/audio/devices/shaping.test.ts:161:30
     expect(brightness(pushed)).toBeGreaterThan(brightness(clean));
```

**It cannot be caused by this branch.** `git diff --name-only d9cc87e..HEAD` is
`src/editor/InstrumentArea.tsx` and `src/editor/InstrumentArea.test.tsx` — no
file under `src/audio/` is touched. And the same commit `b4dee76` ran this
identical check twice: job `95689263238` failed, job `95689285633` **passed**.
Same code, same workflow, opposite outcomes.

**What is actually going wrong.** I measured the brightness ladder that
assertion rests on, six consecutive renders, and locally it is exactly
reproducible:

```
drive=0    brightness=0.0296  rms=0.60432  hf=0.01787
drive=0.5  brightness=0.0656  rms=0.29648  hf=0.01945
drive=1    brightness=0.1196  rms=0.15697  hf=0.01877
```

Now read the CI numbers against that ladder. `pushed` was `0.0656` — the healthy
drive=0.5 value, to five decimal places. But `clean` was `0.1250`, which is not
the drive=0 value (`0.0296`, 4.2× smaller) — it sits up beside the **drive=1**
rung. The drive=0 render came back measuring like a driven one.

So this is not measurement noise on a marginal threshold: the ladder has 2.2×
between each rung and is bit-identical run to run. One render observed another
render's signal. `render()` in that file builds a device node inside
`Tone.Offline` and connects it with `node.output.toDestination()`, and **never
disposes it** — only one test in the file calls `dispose()`. `Tone.Offline`
swaps a global context, so an undisposed graph from a previous render is the
obvious candidate for what bleeds in when the renders are not cleanly
separated, which is why it surfaces on a loaded CI runner and never on an idle
laptop. I have proved the bleed, not the exact route it takes.

That is a real defect in the FX test harness, not in the FX code, and it will
keep reddening unrelated PRs until the harness disposes what it builds. It
belongs to whoever owns `src/audio/devices/`, so I have not touched it here —
weakening or re-running that assertion would be the wrong fix, and it is
outside this PR's purpose. Happy to fix it properly under its own issue if
that is wanted.

The rest of the matrix on `b4dee76` — production build, emulator tests, and the
Chromium/Firefox/WebKit browser suites — is green.

---

Issue #251, opened 2026-08-06
Flaky: shaping.test.ts "adds harmonics as drive climbs" fails intermittently in the full suite


afternoon:
`src/audio/devices/shaping.test.ts › overdrive and saturator (FX-01, FX-02) ›
adds harmonics as drive climbs, from coloration to destruction` fails
intermittently when the whole unit suite runs, and passes when the file runs on
its own. It is red on identical code that was green minutes earlier, so it costs
a CI round whenever it lands.

```
AssertionError: expected 0.06561431610569629 to be greater than 0.11960337447199143
 ❯ src/audio/devices/shaping.test.ts:161:30
   expect(brightness(pushed)).toBeGreaterThan(brightness(clean));
```

## Evidence that it is non-deterministic, not a real regression

- **The same commit passes on re-run in CI.** On [run 31130546042](https://github.com/afternoon/solid-groove/actions/runs/31130546042)
  (`9d96b92`, a branch whose diff is three editor files) the unit job failed on
  this assertion; re-running the failed jobs on the *unchanged* commit passed it.
- **Locally on that branch:** 4 full-suite runs → 2 passed, 2 failed. The same
  file in isolation (`bun run test -- src/audio/devices/shaping.test.ts`) → 8
  passed, every time.
- **On `main` (`efdad99`) locally:** 5 full-suite runs, 5 passes. So the rate is
  not uniform — it appears to rise with how much else the suite is doing, which
  is why an unrelated branch that merely *adds tests elsewhere* can tip it.

The failure signature is stable across occurrences: `pushed` measures
≈`0.0656` while `clean` measures ≈`0.12`–`0.125`. The driven render is coming
out *less* bright than the clean one, which is backwards — that looks like one
of the three `Tone.Offline` renders returning something other than the settled
steady state, rather than the assertion being marginally too tight.

## Where to look

`render()` does `Tone.Offline(..., 0.4)` three times and `settled()` takes the
second half of each buffer, on the assumption that the smoothing ramp is over by
then. Two candidates worth checking before touching the threshold:

- a short or truncated offline render under load, so `settled()`'s window still
  contains the parameter ramp rather than steady state;
- `hfEnergy` being a first-difference RMS, so it scales with sample rate — if a
  render ever comes back at a different rate than its sibling, the comparison
  between two renders is no longer like-for-like.

Same family as #203 (`testAudioTeardown.test.ts` flaking under parallel
worktrees): audio tests that are sound in isolation and fragile under
[...]

PR #264, opened 2026-08-18
Fix #225 (1/1): Keep the sampler's three groups on one row

afternoon:
## What & why

Reported while testing the #225 stack on a preview channel: the sampler panel's
layout is broken — the amp envelope sits on a row of its own with a wide empty
gap beside it, and the panel is roughly twice as tall as it should be.

**Cause: #246 replaced the swap list with prose.** The Sample group now holds
the loaded sound's name and a "Drag a sound here from the library" hint, and
prose sizes itself to its own longest line. Measured in Chromium at the default
1280×720 viewport:

| | Sample group | groups row | panel height |
|---|---|---|---|
| before the stack | 159px | all three on one row | 273px |
| with #246 | **194px** | amp envelope wraps to row 2 | **491px** |

The panel's content box is 616px. The two slider groups need 184 + 242 with a
12px gap, so the Sample group had 166px to play with; at 194px the row overflows
by 28px and the last group wraps, leaving 374px of dead space beside it.

`git log -S` pinned the change to #246's commit — the hint and the sample-name
label arrive together there, and no earlier commit in the stack touches the
group's width.

**The fix keeps the prose out of the wrap calculation rather than shortening
it.** A zero flex-basis means the Sample group contributes nothing to the row's
wrapping decision: the slider groups take the width they need, the sample group
grows into what is left, and its text wraps *inside* that width instead of
setting it. `min-width: 0` is what lets it fall below the hint's own line
length. Copy is untouched, and the panel adapts rather than depending on the
hint's exact wording.

After: one row again, Sample group 166px, panel back to **273px** — the height
it was before the stack.

Refs #225

## Core flows

Untouched. No flow spec, no `docs/core-flows.md`, no `docs/prd.md`.
`bun run verify:core-flows` passes: 3 flows, each with exactly one spec.

## Walkthrough

The change is one panel's layout, so the evidence is the geometry table above
plus the browser assertion below rather than a flow walkthrough — this PR closes
no issue and delivers no new flow. After the fix the panel renders as: SAMPLE
(name, with the hint wrapped to two lines) · PLAYBACK · AMP ENVELOPE, side by
side on a single row, with the Audition button beneath.

## Evidence

A failing test first, in the existing instrument browser spec (not a core-flow
spec):

```
FAIL [chromium] › e2e/instrument.spec.ts › keeps the sampler's parameter groups on one row
  Error: expect(received).toBe(expected)
  Expected: 1
  Received: 2      // two distinct group top-edges: one row became two
```

It asserts the invariant that actually broke — every group shares a top edge —
rather than pinning pixel widths that a font change would falsify.

Re-verified against the current base after `main` absorbed the #225 stack:

```
$ bun run typecheck
(no output)

$ bun run check
Checked 498 files in 1005ms. No fixes applied.

$ bun run test
Test Files  174 passed (174)
     Tests  2343 passed (2343)          # 4 consecutive runs

$ bunx playwright test e2e/instrument.spec.ts --project=chromium
2 passed

$ bun run verify:core-flows
check-core-flows — 3 flow(s), each with exactly one spec.
```

One honest note on those runs: the *first* full-suite run on the rebased tree
exited non-zero, and I did not capture which test it was before it scrolled. The
four runs since have all passed 2343/2343. Given #251 — the offline-render
harness under `src/audio/devices/` returning the wrong or no audio under load,
which has already reddened this PR once with `reverb.test.ts` — that is the
likely culprit, but I did not see it, so I am not claiming it. Nothing in this
diff touches `src/audio/`.

Chromium only locally — this environment cannot reach `cdn.playwright.dev`, so
Firefox and WebKit are not installed. CI runs the full matrix on this branch and
has been green on it, Firefox and WebKit included.

## One deliberate change worth a reviewer's eye

`src/monitoring/replayPrivacy.test.ts` pins the literal
`class="instrument-panel-group"` in `SamplerPanel.tsx`, so that re-broadening
session-replay masking over that container — which would mean rewriting the
attribute as a template literal carrying `MASK_CONTENT` — fails the guard. The
sample group now also carries a layout class, so the pin is updated to
`class="instrument-panel-group sampler-sample-group"`.

The guard's meaning is unchanged: it still asserts a plain, unmasked string
literal on that element, and it still fails if masking creeps back over it. The
diff shows the attribute plainly unmasked. Flagging it because it is a privacy
guard and a pin update should never pass unnoticed.

## Status

Rebased along with the rest of the #225 stack, and now based on `main`, which
has absorbed that whole stack (#242 → #246, plus #262's drop-target fix). This
PR is the last piece of that work still outstanding: `main` carries the drag and
the drop, but not this layout repair.

It closes no issue, so it carries `Refs` rather than `Closes`. Folding it into
#246 is no longer an option — that PR has merged — so it stands on its own.

### DIFF
--- a/e2e/instrument.spec.ts
+++ b/e2e/instrument.spec.ts
@@ -15,17 +15,19 @@ test("changes a track's instrument from its own panel", async ({ page }) => {
 	await page.getByRole("button", { name: "New Project" }).click();
 	await expect(page.getByRole("region", { name: "Step editor" })).toBeVisible();
 
+	// `exact` because the track's instrument controls now sit in a region named
+	// "<track> instrument" (#225), which a substring match would also select.
+	const picker = page.getByRole("region", { name: "Instrument", exact: true });
 	// The workspace scrolls inside a viewport-height app, so each capture
 	// scrolls that container and then returns the page itself to the top —
 	// otherwise the shot is framed on empty page below the editor.
-	const picker = page.getByRole("region", { name: "Instrument" });
 	const show = async (name: string | RegExp): Promise<void> => {
 		await page.getByRole("region", { name }).scrollIntoViewIfNeeded();
 		await page.evaluate(() => window.scrollTo(0, 0));
 	};
 	// The starter track is a sampler, and the panel now says so out loud.
 	await expect(page.getByRole("region", { name: "Sampler" })).toBeVisible();
-	await show("Instrument");
+	await show(/^Instrument$/);
 	await step("Open a project: the track's instrument is a sampler");
 
 	// Sampler -> synth, in one transaction the command layer can name.
--- a/e2e/instrument.spec.ts
+++ b/e2e/instrument.spec.ts
@@ -58,3 +58,23 @@ test("changes a track's instrument from its own panel", async ({ page }) => {
 	await show(/^Drum machine/);
 	await step("Undo returns the drum machine, one switch at a time");
 });
+
+// #246 replaced the sampler's swap list with the loaded sample's name and a
+// "Drag a sound here from the library" hint. That hint sets the Sample group's
+// width from its own longest line, which pushed the three parameter groups past
+// the panel's content box: the amp envelope wrapped onto a second row and the
+// panel grew from 273px tall to 491px, with a wide empty gap beside the wrapped
+// group. The groups belong on one row.
+test("keeps the sampler's parameter groups on one row", async ({ page }) => {
+	await page.goto("/dashboard");
+	await page.getByRole("button", { name: "New Project" }).click();
+	await expect(page.getByRole("region", { name: "Sampler" })).toBeVisible();
+
+	const groups = page.locator(".sampler-panel .instrument-panel-group");
+	await expect(groups).toHaveCount(3);
+	const tops = await groups.evaluateAll((elements) =>
+		elements.map((element) => Math.round(element.getBoundingClientRect().top)),
+	);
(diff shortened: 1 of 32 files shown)


afternoon:
CI red on `typecheck, check:ci, unit + component tests`, and it is #251 again — a different test this time, so recording it here rather than leaving the red unexplained.

```
FAIL src/audio/devices/reverb.test.ts > reverb (FX-01)
     > rings on after the source stops, and longer with a longer decay
AssertionError: expected 0 to be greater than 0
```

Not this branch's code: the diff is four files under `instrument/`, `e2e/` and `monitoring/`, nothing under `audio/`.

**The same commit passed the identical check in the parallel run.** On `c0a9498`:
[96034142070](https://github.com/afternoon/solid-groove/actions/runs/32241918793/job/96034142070) succeeded,
[96034153738](https://github.com/afternoon/solid-groove/actions/runs/32241922641/job/96034153738) failed. Locally that test passes 3/3 on this tree, and the full suite (2339 tests) passed before the push.

`expected 0 to be greater than 0` means both reverb tails measured exactly zero — both offline renders came back silent. Analysis and the accumulating evidence are on #251; I have not touched the harness here because it is outside this PR's purpose.

Also worth noting on this PR: its branch was force-pushed to #262's head (`649d1c1`) at some point this morning, which replaced the layout fix with #262's commits. I have restored `c0a9498`, and the diff is back to the intended four files. I did not make that push and cannot tell from here what did.

---

PR #271, opened 2026-08-26
Migrate to Solid 2.0 RC (1/1): framework, serving layer, and toolchain

afternoon:
## What & why

Moves Solid Groove from `solid-js@1.9.9` to `solid-js@2.0.0-rc.3`, with no Solid 1 idioms left behind. Landed in four waves, plus a review round.

**This PR deliberately departs from two repo conventions, agreed with the repo owner before starting.** Stating them up front so a reviewer does not have to guess:

1. **The title format does not match `$ACTION #<issue> (i/N)`, and there is no `Closes #`.** This work is tracked by branch and PR only. It delivers no user-visible behavior and has no core flow to contract against, so filing it as a feature issue would have meant inventing a spec for a framework swap.
2. **The 400-line ceiling is waived.** Solid 1 and Solid 2 cannot coexist in one bundle — the JSX transform is global — so no intermediate slice can be green on its own commit. The stacking rules exist to let the front of a stack merge without the back; here there is no such front. The wave breakdown below is the intended reading order instead.

Full dependency ledger, code inventory and wave plan: https://claude.ai/code/artifact/d3f5c30a-b167-45e6-a3c9-963cf089db1b

### The finding that shapes everything

There is no `@solidjs/start` release for Solid 2 — all 102 published versions, `2.0.4` included, pin `solid-js ^1.9.x` and `@solidjs/router <2.0.0-0` — and there is not meant to be one. Start's serving layer is now a mode of `@solidjs/vite-plugin`. So this is not a version bump; it is leaving SolidStart. Recorded as **ADR 0005**.

`solid({ start: true })` (client mode) reproduces exactly what `ssr: false` plus `prerender: { routes: ["/"] }` did: a static `dist/client` whose `index.html` is the shell prerendered once, history fallback for deep links, and `render()` rather than `hydrate()`. The app still never renders on the server, so the PRD's client-only decision is unchanged.

Client mode is marked *experimental* by the plugin. That is a carried risk, accepted rather than discovered — SSR start mode is better supported, but enabling SSR would contradict the PRD.

### Substitutions forced by the framework, not chosen

| Package | Why it could not stay |
|---|---|
| `@sentry/solidstart` → `@sentry/browser` | Both `@sentry/solidstart` and `@sentry/solid` peer Solid 1. Same SDK release (`10.68.0`); the symbol set used is byte-identical before and after, so `sentrySink.ts` remains the one module importing `@sentry/*`. Cost: SolidStart's router-tracing integration, which was never enabled. |
| `solid-icons` 1.1.0 → 1.2.0 | 1.1.0's `IconTemplate` used `splitProps`, `mergeProps`, and a signal write inside an effect — the last throws in Solid 2. 1.2.0 rewrote it as a pure function. |
| `solid-firebase` removed | Its only use was a `FirebaseProvider` wrapping the dashboard whose context nothing read. |

`vite` 6 → 8 is pinned exactly, not loosely: `@solidjs/vite-plugin` peers `vite ^8 \|\| ^9` and `vitest@4.1.11` peers `vite ^6 \|\| ^7 \|\| ^8`. The intersection is a single major. Every Solid package is pinned exactly so an RC bump mid-migration is a deliberate act rather than a lockfile surprise.

### File-based routing has no provider on this stack

`FileRoutes` came from `@solidjs/start/router`. Router 2 does ship a `fileRoutes()` adapter in `@solidjs/router/fs`, but it consumes a `virtual:file-routes` manifest that `@solidjs/vite-plugin` does not emit — I checked the package. The four routes are written down in `src/router.tsx` instead. That is also why the page modules lose their `[id]` / `[...404]` filenames: the bracket syntax addressed `FileRoutes`, and with nothing reading it, it only advertised a convention the app no longer has.

## Three real bugs this surfaced

All the same class, and it is worth naming because it is *not* the one the migration guide leads with. **Effect splits produced no defects at all.** Every bug was a read that used to see a write in the same tick and no longer does — invisible to the type-checker, the linter and the idiom guard alike.

1. **Every piano-roll note drag threw.** Two `pointerup` handlers both ended the drag; Solid 1 made the first's `setDrag(null)` visible to the second synchronously. The throw reached `window.onerror`, which `globalHandlers.ts` reports as `{fatal: true}` — so every user who dragged a note counted as a **crashed session** against the PRD section 11 measure. Fixed with a non-reactive `finished` flag, the pattern `velocityDrag` in the same file already used. Regression test asserts on window `error` events, because jsdom reports a listener throw to the virtual console rather than to the caller.
2. **The library panel opened onto nothing.** `useLibraryBrowser.open()` wrote and then read the same state in one tick, so it resolved no pack and fetched no manifest — silently, with no error anywhere. Fixed by resolving from the value already in hand rather than by adding a flush.
3. **A rejected repository hung the editor for ever** — no error UI, nothing reported, where Solid 1's `createResource` had surfaced it. The effect now carries an `error` arm.

## Core flows

**Untouched.** `CF-001`, `CF-002` and `CF-003` remain at `test.fixme`, and no flow spec's assertions were changed. This is not the work that delivers them — `verify:core-flows` reporting them parked is the correct outcome here, not a regression.

Two Playwright **config** files changed, in both suites: `webServer` now waits on `port` rather than `url`. Client start mode serves the document shell only to HTML-accepting GETs and 404s everything else, so Playwright's bare readiness probe read a working server as down and sat until the 120s timeout. `port` waits on the TCP listener, which is what "the dev server is up" actually means. No spec, assertion, or selector was touched.

## Walkthrough

**No UI change.** Nothing a user sees is different — this replaces the framework underneath an unchanged interface. That the interface is unchanged is exactly what the green browser matrix below asserts.

## Evidence

Every check green on `db79145`, the current head:

| Check | |
|---|---|
| `typecheck, check, unit + component tests` | ✅ 176 files, 2358 tests |
| `browser e2e (chromium / firefox / webkit)` | ✅ all three |
| `browser e2e against the emulator (chromium / firefox)` | ✅ both |
| `firebase emulator tests` | ✅ |
| `production build + no-secrets-in-bundle` | ✅ |
| `deploy preview channel` — build, verify, channel deploy, hosted smoke test | ✅ |

**The preview deploy caught a real defect, and it was this migration's.** `verify:client-config` reported all four Firebase fields "not present in the bundle" on a build whose config was complete and correct: Vite 8's rolldown/oxc minifier writes string literals with **backticks** where esbuild wrote double quotes, and the check matched double quotes only. It would have blocked every preview — and, since it is a `predeploy` step, the production deploy at merge. The matcher now accepts all three quote styles, and requires the four fields inside one window rather than searching for each independently across the whole bundle: a config-less build really does ship `devBackend`'s placeholder `mock-project.firebaseapp.com` in one chunk and an `AIza`-shaped key in another, so the per-field search could have assembled a passing result while the object `initializeApp` receives was empty. The script had no test, which is how it broke silently; it has one now (`scripts/verify-client-config.test.mjs`), and with the pre-fix matcher 4 of its 7 cases fail.

The preview channel is live and its hosted smoke test passes — which is the first evidence that a **client start mode** build actually serves from Firebase Hosting and starts a session, not just that it builds.

`bun run check` now also runs **`scripts/check-no-solid-1.mjs`**, a committed guard with one rule per removed or renamed API, so the idioms cannot creep back. Demonstrated red before green rather than asserted: run against `main`'s pre-migration source it reports 45 hits; against this branch, zero.

### Two audio-test findings, and they are not the same finding

Both involve `Tone.Offline` handing back silence, so it would be easy to file them together. They have different causes and different verdicts, and the difference is the whole point — one is not this PR's, the other is.

**Pre-existing, mitigated: overlap *between* files.** `vitest.config.ts` now sets `fileParallelism: false` on the `audio` project. Twenty files there render through `Tone.Offline`, and the backend is `node-web-audio-api` — a native addon, so its state is shared per *process*, not per worker thread, and Vitest's module isolation does not reach it. Measured: 3 failures in 22 local runs and both of two consecutive CI runs before; 16/16 then 3/3 full runs after. Confirmed **pre-existing** by installing `main`'s actual Solid 1.9.9 / Vite 6 toolchain in a separate worktree, where it fails at a comparable rate. Cost ~11s. Review this on its own merits — it is a test-infrastructure change riding along, not part of the migration.

**This PR's, fixed: comparison *across* renders inside one file.** `compressor.test.ts > applies makeup gain` failed CI on `f28f136`. Not a flake to re-run, and — unlike the above — not pre-existing: unmodified `main` on its own toolchain fails **0 / 30**, this branch **7 / 30**. Not a DSP change either: a clean run produces `flat=2.098432e-1`, `loud=8.354009e-1` on both toolchains, bit for bit, while the failures are `1e-23`, `1e-30`, `NaN` — one render coming back silent. Two candidates were tested and ruled out from both directions (`node-web-audio-api` drifted 2.1.0 → 2.2.0 in its caret range here, but 2.2.0 on `main` fails 0/15 and 2.1.0 on this branch still fails 6/30). What is left is the toolchain's timing, which *is* the migration.

So the fix is the fragility the file already documents. Its first test carries the comment *"Comparing two separate renders is what made an earlier version of this flaky: each offline render builds its graph against the shared global context, so the two were not reliably independent"* — and two of its three render-based tests had already been rewritten to measure within one render. This was the one left comparing two. It now puts both settings on two channels of a single render through `Tone.Merge`. The assertion is unchanged and not weaker: mutate `applyParam(makeup.volume, 0, …)` into the source so makeup is ignored and it fails with `expected 1 to be greater than 3`. 30/30 runs of the file pass afterwards. The three other audio tests that do more than one offline render (`reverb`, `delay`, `MasterAudioGraph`) were stressed 20× each: 0 failures, so there is no second one of these waiting.

## Review

Wave 3 ran four lanes, none reviewing its own code. Two findings were defects in this PR's own safety work, both demonstrated with a working repro:

- The replay-privacy guard (ADR 0002 decision 1) had gone **fail-open** when `classList` removal forced the mixer strip's classes into an array — masking the surface no longer failed the guard. It now asserts the absence of `MASK_CONTENT` on the element rather than the presence of a spelling, mirroring how `MASKED_NAMES` asserts the opposite, and covers all nine surfaces.
- The Solid 1 guard walked only `src/`, so its four *dependency* rules could never fire; a `vinxi` script in `package.json` passed it.

Also added on a reviewer's suggestion: two contract tests in the command layer pinning the invariant that **an edit replaces the project root rather than mutating it**. Both the audio graph and the arrangement repaint now depend on that identity change, and a violation would silently kill both with no throw and no failing assertion.

### Review round (db79145)

Three findings from @afternoon's review, both of whose questions were "check whether this is really needed" — so both were measured rather than argued.

- **`fireEvent` does not flush.** `@solidjs/testing-library` re-exports it verbatim from `@testing-library/dom` and calls `flush()` only inside its own `render`/`cleanup`; a click on a counter leaves the DOM reading the old value until `flush()` runs. So the helper stays, but is now shared — `src/testing/events.ts` owns `clickAndFlush` and a general `fireAndFlush`, replacing **three** private near-duplicates (`ArrangementView`'s `clickAndFlush`, `ShortcutGuide`'s `fireAndFlush`, `InstrumentKindPicker`'s `click`), each of which had grown its own copy of the same explanatory comment. 38 sites across 8 files.
- **Solid will not encode ARIA states for us, and fails dangerously.** `aria-pressed={true}` renders `aria-pressed=""`; `aria-pressed={false}` removes the attribute entirely — so an unpressed toggle stops being a toggle to a screen reader. `new String(...)` (and plain `String(...)`) produce the right value but do not typecheck: Solid types these as `EnumeratedPseudoBoolean = "false" | "true"`. So the two literals are spelled **once**, typed, in `src/shared/aria.ts`'s `ariaBool()`, and the 17 sites call it. Its test pins the Solid behaviour, so a later RC fixing it surfaces as a failing test.
- **`retry` in `Dashboard.tsx` is load-bearing, not dead.** The property assignment *is* the tracked read of `retryCount()` — only the compute half is tracked. Deleting it turns "retries the listing when 'Try again' is clicked" red. The explanation sat twelve lines above the code, which is why it read as vestigial; it is now on the line.

## PRD

**`docs/prd.md` had four now-false statements naming SolidStart. Reported rather than edited at first, since the PRD is read-only to implementers; the product owner has since asked for the correction, so this PR now carries it.**

- §9.1's committed-alpha-stack row now reads `SolidJS 2 and TypeScript, served by @solidjs/vite-plugin's client start mode` and links ADR 0005 — §9.1's own rule is that "a decision that changes a row in the table below updates that row to point at its ADR", the way the error-monitoring row points at ADR 0001.
- §7.10 names `@sentry/browser` instead of "the official SolidStart SDK", and says which clause of ADR 0001 was superseded and which was not: Sentry-the-platform is still committed and still needs its own superseding ADR to replace; only the wrapper changed, for the peer-dependency reason ADR 0001's risk list already anticipated.
- §16's committed-technologies bullet and the "Current prototype baseline" paragraph get the same correction.

**No product behavior is revised.** Client-only rendering, the Sentry commitment and the reporting boundary are all unchanged — the PRD now describes the mechanism that actually implements them. ADR 0005's "does not amend the PRD" note is updated in the same commit, since it asserted a present-tense fact this edit makes false.


afternoon:
## CI is red here by construction — not being chased per-push

**Failing check:** `typecheck, check, unit + component tests` ([run](https://github.com/afternoon/solid-groove/actions/runs/33011233465/job/98317738386)), on `80aef80`.

It dies at `tsc --noEmit` with exit code 2, **before a single test executes** — which is also why the `vitest-report-unit` artifact upload warns that no file was found. That is not a second failure, just the first one's shadow.

This failure is genuinely this PR's diff, not an unrelated or flaky one, so I want to be explicit rather than let it read as unexamined. 82 type errors, all of the same few kinds and all in directories this commit deliberately did not migrate:

```
39 × TS2305  Module '"solid-js"' has no exported member 'JSX' / 'onMount'
30 × TS2322  classList not a property; ARIA/DOM attribute types
 5 × TS2353  store setter shape
 3 × TS2724  no exported member 'batch'
 5 × other
```

`src/app.tsx`, `src/router.tsx` and `src/routes/` typecheck clean. The rest is wave 1's, split per-directory: `editor` 36, `library` 17, `components` 12, `instrument` 8, `auth` 4, `shortcuts` 3, `arrangement` 2.

**Why there is no fix to push.** The fix *is* the remaining migration. Solid 1 and Solid 2 cannot coexist in one bundle — the JSX transform is global — so there is no intermediate commit where this check passes. Going green means completing waves 1–3, which is the whole job, not a patch to this one. The repo owner and I agreed before starting that CI would be red until then and would be checked once, as the full three-browser matrix, before this leaves draft. I am not re-running the job: this is a real, reproducible, expected failure, and "flake" would be the wrong word for it.

The wave-0 gate this commit was measured against is the dev server, not CI — `bun run build` also fails by design, because a production build bundles every route while the dev server compiles lazily. That asymmetry is precisely what lets the landing page boot on Solid 2 while the editor is still Solid 1, and it is what makes the incremental plan work at all. Evidence for the gate is in the PR body.

### What this run usefully caught

Two DOM-attribute changes I had under-weighted when I inventoried the work, both from the migration guide's "attributes are attributes, and generally lowercase" rule:

- **Boolean-valued ARIA attributes are now type errors** — `aria-pressed={bool}` (16 sites) and `aria-expanded={bool}` (2). They want the string form. `aria-label={expr}` and friends are string-valued already and unaffected, so the real count is 18, not the 51 a naive `aria-*={` grep suggests.
- **camelCase DOM attributes** — `tabIndex` → `tabindex`, `maxLength` → `maxlength`. Two sites.

Small, but they are spread across the `library/` and `instrument/` surfaces rather than concentrated, so they are going into the wave-1 briefs rather than being discovered five times over.

I will not comment again on this check while this blocker holds; the next update here is wave 2 turning it green.

---

src/components/Dashboard.test.tsx
```diff
@@ -202,8 +206,10 @@ describe("Dashboard", () => {
       renderDashboard();
 
       fireEvent.click(await screen.findByRole("button", { name: /rename/i }));
+      flush();
```
afternoon: Let's extract the clickAndFlush helper to a test helpers file and reuse it across any test that clicks and then flushes.

src/components/Dashboard.tsx
```diff
@@ -49,45 +50,60 @@ export default function Dashboard(props: DashboardProps = {}) {
...
+  // where a write is sanctioned, and the cancellation that used to be a
+  // nested `onCleanup` is the cleanup that half returns.
+  //
+  // `prev === undefined` is how "first run" is spelled now that
+  // `createEffect`'s second argument is the apply function rather than 1.x's
+  // `initialValue`.
+  createEffect<ProjectsListing>(
+    (previous) => ({
+      id: userId(),
+      retry: retryCount(),
```
afternoon, line 71: `retry` isn't used.

src/components/ProjectList.test.tsx
```diff
@@ -81,8 +93,10 @@ describe("ProjectList", () => {
     renderWithRouter([project], { onRename });
 
     fireEvent.click(screen.getByRole("button", { name: /rename/i }));
+    flush();
```
afternoon: use `clickAndFlush` helper

src/editor/Mixer.tsx
```diff
@@ -308,7 +303,7 @@ function TrackStrip(props: TrackStripProps): JSX.Element {
         <button
           type="button"
           class="mixer-strip-select"
-          aria-pressed={props.selected}
+          aria-pressed={props.selected ? "true" : "false"}
```
afternoon: I'm seeing this pattern a lot. Is it really necessary? Hardcoding the string values everywhere seems error prone. Replace with `new String(props.selected)` instead.

src/shortcuts/ShortcutGuide.test.tsx
```diff
@@ -44,6 +45,18 @@ function renderGuide(
   return { transport, onClose };
 }
 
+/**
```
afternoon: This method can go in a test helpers module as well.

src/shortcuts/ShortcutGuide.test.tsx
```diff
@@ -160,8 +179,10 @@ describe("search and context filter", () => {
   it("filters to what is available right now on request", () => {
     renderGuide({ contexts: ["editor"] });
 
-    fireEvent.click(
-      screen.getByLabelText("Only shortcuts available here", { exact: false }),
+    fireAndFlush(() =>
```
afternoon: use `clickAndFlush` for legibility.

afternoon (commented):
- There are a lot of click-then-flush points in the code.  Check that the fireEvent​ call doesn't already flush​. If we need to flush ourselves, generalise the clickAndFlush​ helper. That's a good pattern, let's reuse it.
- I see a lot of attribute setters like aria-pressed={props.metronomeEnabled() ? "true" : "false"}​, encoding the string values seems error prone. Do we _really_ need to do this, will Solid not do it? If we _have_ to, lets do `aria-pressed={new String(props.metronomeEnabled())}` instead, which doesn't rely on us spelling both string values correctly. These seem to be prevalent on aria-* attributes.

src/components/Dashboard.tsx
afternoon: It reads that way, but it is load-bearing — the property assignment *is* the tracked read of `retryCount()`. Only the compute half is tracked in Solid 2, so this is where the dependency has to be registered; the apply half then never needs the value, which is why it isn't destructured.

Confirmed rather than assumed: deleting the line turns `Dashboard.test.tsx`'s "retries the listing when 'Try again' is clicked" red — the button silently stops retrying.

The explanation existed, but twelve lines above the code, which is exactly why it read as vestigial. Moved onto the line itself in db79145.

---

src/editor/Mixer.tsx
afternoon: Good question, and I measured it rather than guessing. **Solid will not do it**, and the failure mode is worse than untidy — measured against `solid-js@2.0.0-rc.3`:

| written | rendered |
|---|---|
| `aria-pressed={true}` | `aria-pressed=""` |
| `aria-pressed={false}` | attribute removed entirely |

Solid treats it like `disabled`, a present/absent HTML boolean. But these are *enumerated string* attributes: `"false"` is meaningful and is not the same as absent. A removed `aria-pressed` means "not a toggle button at all", so a screen reader stops announcing the pressed state — a real accessibility defect, not cosmetics.

On `new String(...)`: it does produce `"false"` at runtime, but it doesn't typecheck, and neither does plain `String(...)`. Solid types these as `EnumeratedPseudoBoolean = "false" | "true"`, and `string` isn't assignable to that:

```
Type 'string' is not assignable to type 'RemoveAttribute | EnumeratedPseudoBoolean | "mixed"'
```

So I took your intent — don't rely on spelling both literals at 17 sites — and spelled them once instead, typed: `ariaBool()` in `src/shared/aria.ts`, returning `"true" | "false"`. All 17 sites now read `aria-pressed={ariaBool(props.selected)}`. It also avoids the boxed-`String`-object footgun, which stays a wrapper object right up until something reads it.

Its test pins the Solid behaviour above, so if a later RC starts encoding these properly we find out from a failing test rather than by noticing.

db79145.

---

src/components/Dashboard.test.tsx
afternoon: Checked the premise first: `fireEvent` does **not** flush. `@solidjs/testing-library` re-exports it verbatim from `@testing-library/dom` (`export * from "@testing-library/dom"`) and calls `flush()` only inside its own `render` and `cleanup` — and measuring agrees, a click on a counter leaves the DOM reading the old value until `flush()` runs. So the helper is needed.

Extracted to `src/testing/events.ts` as `clickAndFlush`, plus a general `fireAndFlush(fire)` for non-click events. That also collapses three private near-duplicates that had each grown their own copy of the explanation: `ArrangementView`'s `clickAndFlush`, `ShortcutGuide`'s `fireAndFlush`, and `InstrumentKindPicker`'s `click`. 38 click-then-flush sites across 8 files now go through it.

I left `fireEvent.input(...); flush();` and friends as two lines — `fireAndFlush(() => fireEvent.input(...))` reads worse than what it replaces — but the general form is there when a callback is the right shape.

db79145.

---

src/components/ProjectList.test.tsx
afternoon: Done — all 8 click-then-flush sites in this file now use the shared `clickAndFlush` from `src/testing/events.ts`. db79145.

---

src/shortcuts/ShortcutGuide.test.tsx
afternoon: Moved — it is now `fireAndFlush` in `src/testing/events.ts`, the general form alongside `clickAndFlush`. This file imports both; the three search-filter sites keep the callback form, which is what it is for. db79145.

---

src/shortcuts/ShortcutGuide.test.tsx
afternoon: Done — this one is now `clickAndFlush(screen.getByLabelText(...))` instead of `fireAndFlush(() => fireEvent.click(...))`, which drops two lines of wrapping. db79145.

---

Issue #283, opened 2026-08-27
LOOP-020 - The master FX panel in the mixer


afternoon:
> **Respecced by #304.** This issue used to own two things: a switchable main
> region (Arrangement | Master) and the master FX panel. #304 replaces the whole
> editor shell with three views and a dock, so the view-switching half is gone —
> and the master chain moves into the **Mixer** view rather than a tab. What is
> left here is the panel itself. The acceptance criteria below were rewritten on
> that basis; the ticks they used to carry did not correspond to any code on
> `main`.

## Scope

The **master device chain, in the Mixer view**: selecting the master strip
reveals the master's chain, an "+" that adds one of the six device types, and
the selected device's controls.

The DSP already exists — `LOOP-009` (#49) shipped `filter`, `overdrive`,
`saturator`, `compressor`, `delay` and `reverb` with real processing, and
`device.add/remove/reorder/duplicate/setBypass/reset` are registered commands.
**This task is the UI.**

#304 leaves the mixer's master strip as the route to this panel, so this task
adds the panel behind a place that already exists rather than inventing an
entrypoint.

**Core flow: CF-007**, rewritten by #304 for the new shell — it walks adding an
**overdrive** to the master and driving it while the loop plays.

## Why

Device chains are unreachable in the product today (#241 tracks the same gap for
track chains): a producer can build a beat and cannot put a single effect on it.
The master is the right first chain — it always exists, needs no selection model,
and "glue the whole thing" is the move that most changes how a loop sounds.

It is also where the chain UI gets built *once*: #241 then reuses the same
components for a track's inserts in the Instrument view, whose slot #304
reserves.

## PRD

- **Principle 5, legible controls** — every device exposes its musically relevant
  controls, spelled out, consistent in shape, with the effect on the sound shown
  where it can be drawn.
- **Principle 10, make room for the unexpected** — broad parameter ranges,
  deliberate misuse reachable and labelled, no silent normalisation toward a
  conservative sound. The master safety limiter protects the output without
  rewriting stored parameters.

## Acceptance criteria

- [ ] Selecting the master strip in the mixer reveals the master's device chain.
- [ ] The panel lists the master chain in order and has an add control offering
      the six registered device types **from their registry definitions** — no
      hard-coded device list in the component.
- [ ] Adding, removing, reordering, duplicating, bypassing and resetting a master
      device each go through the existing commands: one revision and one history
      entry apiece, undoable.
- [ ] A device's controls are generated from its parameter definitions (range,
[...]

Issue #282, opened 2026-08-27
LIB-09 - Personal packs: create a pack and add sounds by dropping files onto it


afternoon:
## Scope

A signed-in user can make their own pack in the library browser and fill it by dragging audio files from the operating system onto it.

- **Add pack** button creates a pack row with its name in an inline input; Return commits the name.
- Dropping files on **empty space** in the user-library area creates a pack called **"My Sounds"** and imports into it.
- Dropping files on an **existing** user pack imports into that pack.
- Imports are first-class library entries: searchable, filterable, auditionable, draggable onto a track exactly like factory content.
- **Sign-in is required.** A guest who tries gets the existing upgrade-account prompt.
- **500 MB per free account**, enforced client-side and server-side.

This is the alpha slice of #120 (P1-005 / LIB-03); #120 remains the umbrella for the rest of the personal library.

**Core flow: CF-006.**

## Why

Bringing your own sounds is the difference between a toy and an instrument, and the pack model already exists to hold them — LIB-03 requires imports to land in the user's own pack rather than in a parallel content model, so one browser, one resolver, one audition path and one export path serve factory and user content alike.

LIB-03 explicitly leaves open "whether a user can create more than one personal pack, and how those packs are named and organized". This task settles it: **many packs, user-named, created explicitly or implicitly by a drop.**

## PRD

- **LIB-03** — user sample imports and personal library (the whole requirement; note the criteria on progress, rejection messages, rename/delete, per-user scoping, and server-side enforcement).
- **LIB-05** — pack identity, versioning, and `kind: "user"`.
- **LIB-01** — imported assets behave like curated ones for search, filter and audition.
- **Security and privacy** (§10) — per-user scoping, safe content headers, and no redistribution of user audio.

## Acceptance criteria

- [ ] "Add pack" creates a `kind: "user"` pack owned by the signed-in account, with the name in an editing input until Return commits it. Clicking away without pressing Return keeps a default name rather than losing the pack.
- [ ] A drop on empty space creates a pack named "My Sounds" (renameable afterwards) and imports the dropped files into it.
- [ ] A drop on an existing user pack imports into that pack. Multi-file drops are supported. Drop targets show an affordance on drag-over and accept only supported types.
- [ ] Each file shows its own progress and a clear success, failure, or cancel state. A failed or cancelled upload leaves no partial or unplayable asset behind.
[...]

Issue #281, opened 2026-08-27
LOOP-019 - Insert a library loop to get a new track


afternoon:
> **Respecced by #304.** The library becomes a modal opened from a slot, so
> there is no longer a panel to drag a sound out of — and nothing to drop it
> onto, since the modal covers the arrangement. The outcome this issue promised
> is unchanged: a loop out of the library lands on a new track at bar 1 and
> nothing else in the project moves. What changes is the gesture. **There is one
> insert action, and the asset's kind decides what it does** — the same button
> that loads a sample onto a sampler creates a loop's track.

## Scope

**Inserting a loop** from the library creates a new track at the bottom of the
track list, carrying that loop as a clip at bar 1.

The library is reached from the arrangement (an affordance this task owes, since
#304 removes the always-open panel) and from a sampler's sample slot. The asset's
kind decides what the one insert action means:

- **loop** → always a new track at the bottom, clip at bar 1. No existing track
  is modified, whichever one is selected.
- **one-shot / sample** → unchanged from today (#225, #263): it loads onto the
  edited track's instrument.

Inserting changes **nothing else**. It does not set the song tempo, it does not
touch the loop brace, and it does not start playback.

**Core flow: CF-005**, rewritten by #304 around this gesture.

## Why

Getting a loop in is the first move of the workflow this product is named after,
and today there is no way to do it — the insert path only feeds a sampler's
instrument slot.

One action with a kind-dependent outcome is what the modal library makes
possible, and it is the better shape anyway: a producer learns a single "put this
in my project" control rather than two gestures whose difference is where they
release the mouse, and it is keyboard-reachable for free, which the drag never
was without a parallel path.

Everything the insert *doesn't* do is deliberate. An insert that retempos the
song rewrites work the producer already has; one that moves the brace destroys a
loop range they set on purpose; one that starts playback takes the transport away
from them. The loop follows the project tempo by time-stretching, which is what
INS-02 already requires, so a 140 BPM loop inserted into a 120 BPM project simply
plays at 120 with its pitch intact.

## PRD

- **INS-02** — curated tempo-labelled audio loops that follow project tempo; the
  UI distinguishes a time-stretched loop from a pitched one-shot.
- **LIB-01** — auditioning and loading library content; a missing or corrupt asset
  is isolated and reported.
- **LIB-05** — pack-qualified asset identity; the project records the pack and
  version it resolved from.
- **TRK-01** — track management.

## Acceptance criteria

[...]

Issue #280, opened 2026-08-27
LOOP-018 - Loop brace on the timeline ruler


afternoon:
## Scope

Put the loop brace on the arrangement's timeline ruler and let the producer drag it, plus a loop toggle that decides whether the transport obeys it.

- The brace **always exists** on the ruler — there is no "no loop set" state.
- The toggle is separate from the brace, and is on by default on a new project.
- Nothing in the product ever moves the brace on the user's behalf.

Both values are persisted project state; that is #278, which lands first. This task is the control.

**Core flow: CF-004.**

## Why

A loop you can see and grab is how a producer works on eight bars for an hour. Keeping the toggle separate from the brace is what lets someone let the song run past the loop without losing the range they set — the Ableton behaviour that makes the brace safe to leave lying around.

## PRD

- **AUD-02** — the user can define an arrangement loop range aligned to bars; looping has no accumulating drift or duplicate triggers at the boundary.
- **AUD-01** — the playhead follows audible playback and returns to the expected position on stop.
- **KEY-01/KEY-02** — any key combination comes from the shortcut registry, never from a component's own listener.
- **Legible controls / keyboard and accessibility** (§8) — the brace is a real control, not a decoration.

## Acceptance criteria

- [ ] The brace is drawn on the ruler strip in `src/arrangement/canvasRenderer.ts`, visually distinct from section markers and the playhead, in the same paint pass rather than as an overlay element.
- [ ] Dragging either handle resizes the brace; dragging its middle moves it. Both snap to whole bars via `barAlignedLoop`, and the range can never be empty or inverted.
- [ ] Each drag commits as **one** history entry and one revision (`history.beginGesture()`), while updating the transport continuously as it moves.
- [ ] Dragging during playback keeps playback running; the new bounds are obeyed from the next pass, with no duplicate trigger or drift at the boundary.
- [ ] The loop toggle is reachable and unambiguous, shows its state, and is keyboard operable. Its key comes from the shortcut registry, with the action ID added to `SHORTCUT_ACTION_IDS` and `src/analytics/catalog.ts`, and `docs/shortcuts.md` updated in the same PR.
- [ ] There is a keyboard path to move and resize the brace itself, not only a mouse one, and it is announced to assistive technology. Because the ruler is a canvas, that announcement is also the only thing a test or a screen reader can read: expose an accessible mirror of the current range in inclusive bar wording ("bars 1 to 2"), and keep it the single source of that text.
[...]

PR #284, opened 2026-08-27
Register core flows CF-004 to CF-007

afternoon:
## What &amp; why

Four new core flows and their `test.fixme` specs, written from a product-owner conversation about the Ableton habit they come from: open a pack, drop a loop, play it in a loop, add elements, then play with effects. Each is a journey through behaviour the PRD already specifies — **no PRD change, and no existing flow altered**.

| Flow | Journey | Issue |
| --- | --- | --- |
| CF-004 | The loop brace on the ruler, and the toggle that decides whether the transport obeys it | #280 |
| CF-005 | Dropping a library loop makes a new track at the bottom, clip at bar 1 | #281 |
| CF-006 | Creating a personal pack and filling it by dropping audio files onto it | #282 |
| CF-007 | Switching the main region to the master and driving the mix through an overdrive | #283 |

Supporting issues filed alongside them, not delivered here: #277 (one home for core-flow specs), #278 (persist the loop range and toggle), #279 (resolve pack availability per asset).

Refs #280, #281, #282, #283

### The decisions these flows encode

- **Nothing moves the brace but the producer.** An earlier sketch had a drop set the brace and start playback; both were dropped as too destructive of work already done. CF-005 asserts the *absence* of those consequences explicitly.
- **A drop never retempos the song.** The loop follows the project tempo by time-stretching, which is what INS-02 already requires — so the flow stays inside the PRD rather than introducing a "first loop claims the tempo" rule the PRD does not have.
- **Asset kind decides what a drop means.** A loop makes a new track; a one-shot still loads onto the track under the pointer, as #225/#263 built it.
- **Every flow ends after a reload.** Persistence is part of a flow's outcome now, which is why all four live in `tests/e2e/emulator/flows/` — the mock backend is a fresh empty store on every page load and cannot prove it.

## Core flows

**Registered and specced.** Four entries in `docs/core-flows.md` and four `test.fixme` specs, landing together as the rule requires — a register edit on its own cannot merge green.

```
$ bun run verify:core-flows
  note: CF-004 … CF-007 is still test.fixme in tests/e2e/emulator/flows/… — in flight, or abandoned mid-stack.
check-core-flows — 7 flow(s), each with exactly one spec.
```

No existing flow's assertions were touched: the diff against `main` is insertions only in `docs/core-flows.md`, above the "new flows go here" marker, and four new spec files. CF-001–003 are untouched.

**One flow was corrected while its spec was being written** (`b37c475`), which is the point of writing them together. CF-007 originally read "drive it up → undo once, the overdrive comes off". That cannot be true: a parameter gesture is its own history entry, so one undo would have taken back the drive and left the device on the chain, and redoing the add would have restored it as its payload described it. The undo and redo now come immediately after the add, and the drive is pushed afterwards. Both still precede the reload, because history is session-local. Noted on #283.

## Walkthrough

No UI change — documentation and skipped specs only.

## Evidence

```
$ bun run check            # tsc --noEmit && biome check && verify:no-solid-1 — pass
$ bun run verify:core-flows # 7 flows, each with exactly one spec — pass
$ bun run test:browser:chromium
  2 skipped
  21 passed (34.2s)        # unchanged from before this branch
```

Chromium-only for the browser suite, per `CLAUDE.md` — this environment cannot reach `cdn.playwright.dev` for Firefox and WebKit, and CI gates those on push. The four new specs are `fixme`, so they are skipped everywhere until the issues that deliver them remove the markers.

## Notes for the reviewer

- **Title format.** `CLAUDE.md` defines `$ACTION #<issue> (i/N): Title` for feature and fix PRs; this is the product owner's contract PR, which lands separately from the stacks it measures. Renaming is fine — it belongs to four issues, not one.
- **`blocked_by` still needs setting on GitHub.** The graph is the authority and it cannot be edited through the tools available here. The edges these issues assume: #280 ← #278, #277 · #281 ← #277 · #282 ← #279, #277 · #283 ← #277.
- **Milestones and Projects Status are deliberately unset**, so an unattended `solid-groove-phase-1` run cannot pick any of these up.
- **Three things writing the specs surfaced, all folded back into the issues rather than worked around:** #280's default brace was ambiguously worded (it is one bar, now corrected, with an accessible mirror of the range required because the ruler is a canvas); #281 needs the library row to state a loop's source tempo, or CF-005 step 2 is not a thing a person can do; #282 must add the Storage emulator to `test:browser:emulator`, which currently starts `--only firestore,auth`.
- **One stale line in `CLAUDE.md`:** it says schema v1 registers no concrete processors, but `LOOP-009` (#49) shipped all six with real DSP and registered commands. Worth a one-line fix in a separate PR.

### DIFF
--- a/docs/core-flows.md
+++ b/docs/core-flows.md
@@ -214,6 +214,130 @@ its clips moving with it, and a single undo puts the order back.
 automation (`ARR-004`); and whether sections survive a reload, which would belong
 to a flow in `tests/e2e/emulator/flows/`.
 
+### CF-004 — A producer sets the span they are working in
+
+**Issue:** #280 · **Suite:** `tests/e2e/emulator/flows/CF-004.spec.ts` · **Entrypoint:** the
+project dashboard
+
+**Preconditions:** signed in with no projects.
+
+1. Create a new project. Above the tracks, the ruler carries a loop brace spanning
+   the first bar, and looping is on.
+2. Start playback. The playhead runs to the end of bar 1 and jumps back to the
+   start, over and over.
+3. Drag the right-hand edge of the brace out to the end of bar 2.
+4. The brace now spans two bars. Playback never stopped, and the playhead now
+   turns around at the end of bar 2.
+5. Switch looping off. The playhead runs past the end of the brace and keeps
+   going; the brace stays where it is.
+6. Switch looping back on, stop, and reload the page.
+7. The project reopens with the brace still spanning bars 1 and 2, and looping
+   still on.
+
+**Outcome:** the producer chose the span they are working inside, changed it
+without interrupting playback, and found it exactly as they left it when they
+came back.
+
+**Out of scope:** that any of it is *audible* — a headless browser records no
+audio, so this proves the playhead and the transport, not that a sound reached a
+speaker (see [`docs/testing.md`](./testing.md#playback-is-asserted-in-chromium-only--a-known-tracked-gap)
+and issue #43). It also does not prove that anything else in the product moves
+the brace: nothing does, deliberately, and CF-005 is where that is asserted from
+the other side.
+
+### CF-005 — A producer drops a loop from the library onto a new track
+
+**Issue:** #281 · **Suite:** `tests/e2e/emulator/flows/CF-005.spec.ts` · **Entrypoint:** the
+project dashboard
+
+**Preconditions:** signed in with no projects. The library contains a
+tempo-labelled loop whose source tempo is not the tempo a new project opens at.
+
+1. Create a new project. It opens on the starter kick pattern.
+2. In the library browser, find a drum loop that was recorded at a different
+   tempo from the project's.
+3. Drag it out of the browser and drop it on empty space in the track area.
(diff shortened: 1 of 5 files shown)


afternoon:
CI is red on `ee3a358`, and it is this PR's — expected, and named in the description.

The failing step is `bun run verify:core-flows`, and it is the only failure in the job: `tsc --noEmit`, `biome check` and `verify:no-solid-1` all passed on the same run.

```
check-core-flows — 4 problem(s):
  - CF-004 ... is registered in docs/core-flows.md but has no spec.
  - CF-005 ... is registered in docs/core-flows.md but has no spec.
  - CF-006 ... is registered in docs/core-flows.md but has no spec.
  - CF-007 ... is registered in docs/core-flows.md but has no spec.
```

That is the check doing its job: a register edit on its own cannot merge green, which is precisely what forces a flow and its `test.fixme` spec into one diff.

The fix is the four specs in `tests/e2e/emulator/flows/`, being written on this branch now. No re-run is warranted — the failure is deterministic and will stay red until they land.


---

PR #285, opened 2026-08-27
Fix #258 (1/1): Arrangement shortcuts go dead while the piano roll is showing

### LINKED ISSUE #258: Deleting fails when moving between tracks
Repro:

- Add 2 tracks in a project
- Duplicate clips 3 times for a total of 4 clips
- Select a clip placement in track 2
- Press delete keyboard shortcut to delete it
- Select clip placement in track 1
- Press delete keyboard shortcut to delete it

Expected:

- Selected clip placements from track 1 and track 2 are deleted, no others

Observed:

- First selected clip placement is deleted
- 2nd selected clip placement is not deleted - pressing delete does nothing

afternoon:
## What &amp; why

Selection-scoped shortcuts asked the wrong question. `hasArrangementSelection()` was defined as `!showPianoRoll() && arrangementEditingActions()?.hasSelection()`, and `edit.delete` branched on `showPianoRoll()` before consulting any selection. But `showPianoRoll()` says which editor is mounted *below* the arrangement — not which surface the user selected in, and the arrangement is on screen either way. So a placement selected while a piano roll was showing got routed to a piano roll with nothing selected, and the handler reported itself disabled.

The fix replaces the visibility test with one `selectionOwner()` accessor (`"piano_roll" | "arrangement" | "step_editor" | null`) that every selection-scoped mapping reads. Precedence is unchanged wherever it existed: the piano roll wins while it is showing **and** holds a note selection; only the case where it had no selection changes.

**The reported trigger is not the real one.** The issue describes moving between tracks, but track switching is incidental — clicking a placement re-points the *edited* track (#228), and Delete only breaks once that track is one the editor shows the piano roll for (a synth track holding a note clip). Verified both directions before touching source: two sampler tracks with a delete on each **passes**; one synth track with one placement and no track switching at all **fails**.

Closes #258

## What this PR contains, after a deliberate narrowing

Three defects, all confirmed fixed on the preview channel by the product owner:

1. **Delete, Duplicate, Cut and Copy** reach the arrangement's selection while a piano roll is showing. (The original defect.)
2. **Paste is no longer gated on which editor is mounted.** Its `isEnabled` carried the same `!showPianoRoll()` term, and because a disabled mapping leaves the browser default alone, it failed *silently*. The registry already lists `edit.paste` under both `selection` and `arrangement` (`registry.ts:209`), so only this gate stood in the way. Nothing is stolen from the piano roll: `PianoRollActions` is delete, duplicate, select-all and has-selection, so `Mod+V` has one meaning in the editor.
3. **Only one arrangement selection is painted at a time.** The shell's ARR-01 bar range (`canvasRenderer.ts:278`) and the ARR-002 placement selection (`canvasRenderer.ts:294`) are drawn in the same cyan, and each click set one without clearing the other — a placement hit returns before `shell.handlePointerDown` runs, and an empty-space click never touched the placement selection. They are now mutually exclusive. This one **predates the PR** (`git diff origin/main` on those files was empty) and is included because it is three one-line writes and because two lit boxes make every selection-scoped operation unreadable.

## What was removed, and why

**Paste *anchoring* was attempted here and has been taken out.** `placementEditingController.ts` is byte-identical to `main` again.

The sequence, recorded because it is the useful part: paste ignored the selection and went to the playhead → anchoring it at the selection fixed that but stacked a copy invisibly on its own source → dodging the stack fixed that but ignored the bar-range selection the user had just made to point at a target. Three rounds, each fix uncovering the next.

The reason it kept failing is structural, not carelessness: the anchor depends on two things schema v1 does not have — a rule about whether placements may overlap (#290) and overwrite semantics for what a paste lands on (#291). Without those, every anchor rule is a heuristic that surprises somewhere. #291 will rewrite this code regardless.

So paste lands at the caller's anchor (the playhead), exactly as on `main`. That is **worse than what the reporter asked for** and is stated here rather than buried: pasting with a placement selected still ignores that selection. It is predictable, it is not a regression against `main`, and #290/#291 are where it gets settled.

## Core flows

Untouched. A bug fix's contract is its regression tests. `bun run verify:core-flows` passes — 7 flows, one spec each; CF-006/CF-007 remain `test.fixme` exactly as on `main`. No flow spec, `docs/core-flows.md` or `docs/prd.md` was edited.

## Walkthrough

Not captured. The changes are user-visible — five keyboard operations that did nothing now work, and one selection box is lit instead of two — but there is no *visual* change to screenshot: no new component, layout, styling or copy. The regression tests carry the proof instead, asserting against stored project state and the accessible selection mirror.

## Evidence

Four regression tests, each verified red by reverting **only** its own source change:

```
 FAIL  EditorView > deletes a selected placement while the piano roll is showing (#258)
AssertionError: expected [ 'plc_elLyKQlVOiYStrzI-Qz6h' ] to not include 'plc_elLyKQlVOiYStrzI-Qz6h'

 FAIL  EditorView > pastes a copied placement while the piano roll is showing (#258)
- Expected   2
+ Received   1

 FAIL  ArrangementView > keeps exactly one arrangement selection live at a time
AssertionError: expected 'Selected BD, bars 4 to 5' to be 'No selection'
```

In the `ArrangementView` case the assertion above it (`selectedPlacementIds` is `[placementId]`) passes, so the red is specifically "the placement is selected *and* the earlier bar range is still live". A fourth test pins that paste uses the caller's anchor, so the removal above cannot silently drift back.

| Command | Result |
| --- | --- |
| `bun run typecheck` | pass |
| `bun run check` | pass — biome 507 files, `verify:no-solid-1` clean |
| `bun run test:all` | pass — 177 files, 2366 tests (the command CI runs) |
| `bun run verify:core-flows` | pass — 7 flows, one spec each |

Rebased onto `main` after #286 merged; every check re-run on the rebased head.

Diffstat: 6 files changed, 231 insertions(+), 33 deletions(-) — **264 changed lines**, comfortably inside the 400 ceiling after the narrowing (it had reached 370).

Nothing that predates this PR changed: the pre-existing ARR-002 delete test keeps its own inline selection helper and is byte-identical.

**Browser suites did not run locally.** `bun run test:browser:install` cannot reach the Playwright CDN in this container, which CLAUDE.md anticipates. CI on push is the browser check.

## Follow-ups filed

- #290 — placements can silently overlap; no invariant, no defined resolution. Governs the two below.
- #291 — paste should overwrite what it lands on (Ableton parity), which is where paste anchoring belongs.
- #292 — the arrangement's two selection models, now mutually exclusive rather than unified.

## Acceptance criteria met

- Both selected placements delete, and nothing else does — asserted against the stored project.
- Cut, copy, duplicate and paste work on the same surface and selection as delete.
- Exactly one arrangement selection is painted at a time.
- No contract touched: domain, command registry, persistence, projections and `src/selection/` are unchanged.
- No analytics change. `shortcut_used` is logged by `ShortcutController` from the registry entry; the set of firing actions is unchanged apart from now firing when it should.

### DIFF
--- a/src/arrangement/ArrangementView.test.tsx
+++ b/src/arrangement/ArrangementView.test.tsx
@@ -224,6 +224,41 @@ describe("placement editing wiring (ARR-002)", () => {
     expect(selectedPlacementIds(container)).toEqual([placementId]);
   });
 
+  /**
+   * The preview showed two cyan boxes on one track at once: the shell's ARR-01
+   * bar range and the ARR-002 placement selection, each left standing by the
+   * click that made the other. "There should only be a single selection."
+   */
+  it("keeps exactly one arrangement selection live at a time", async () => {
+    const { renderView, placementId } = await setUpEditing();
+    const { container } = renderView();
+    const canvas = interactionCanvasOf(container);
+    const rowY = RULER_HEIGHT_PX + ROW_HEIGHT_PX / 2;
+    const barRangeText = () =>
+      screen.getByTestId("arrangement-selection-live").textContent;
+    const clickAt = (ticks: number) => {
+      firePointer(canvas, "pointerdown", {
+        clientX: ticks * PIXELS_PER_TICK,
+        clientY: rowY,
+      });
+      firePointer(canvas, "pointerup", { clientX: 0, clientY: 0 });
+    };
+
+    // Empty space, three bars along: the shell takes a bar range.
+    clickAt(TICKS_PER_BAR * 3);
+    expect(barRangeText()).toMatch(/^Selected /);
+
+    // The placement now becomes the only selection.
+    clickAt(TICKS_PER_BAR / 2);
+    expect(selectedPlacementIds(container)).toEqual([placementId]);
+    expect(barRangeText()).toBe("No selection");
+
+    // And back out to empty space: the bar range is the only one again.
+    clickAt(TICKS_PER_BAR * 3);
+    expect(barRangeText()).toMatch(/^Selected /);
+    expect(selectedPlacementIds(container)).toEqual([]);
+  });
+
   it("dragging a placement's body moves it, bar-snapped, through the command layer", async () => {
     const { session, renderView, placementId } = await setUpEditing();
     const { container } = renderView();
--- a/src/arrangement/ArrangementView.tsx
+++ b/src/arrangement/ArrangementView.tsx
@@ -436,6 +436,9 @@ export default function ArrangementView(props: ArrangementViewProps) {
       const hit = shell.hitTestAt(x, y);
       if (hit.kind === "placement") {
         const { tick } = shell.pointToArrangement(x, y);
+        // One selection at a time: the placement becomes it, so the shell's
+        // bar range from an earlier click stops being painted alongside it.
+        shell.setSelection(null);
         editing.beginDrag(hit.placementId, hit.handle, tick);
         activePointerId = event.pointerId;
         (event.currentTarget as Element).setPointerCapture?.(event.pointerId);
@@ -445,6 +448,9 @@ export default function ArrangementView(props: ArrangementViewProps) {
       }
     }
 
+    // ...and the other way round: a bar range is the selection now, so the
+    // placement selection it replaces is dropped rather than left on screen.
+    editing?.clearSelection();
     const selection = shell.handlePointerDown(x, y);
     bumpState();
     if (selection) noteFirstUse();
@@ -495,6 +501,7 @@ export default function ArrangementView(props: ArrangementViewProps) {
     if (!shell) return;
     const track = projection().tracks[rowIndex];
     if (!track) return;
+    editing?.clearSelection();
     shell.setSelection({
       trackId: track.id,
       startTick: 0,
--- a/src/arrangement/placementEditingController.test.ts
+++ b/src/arrangement/placementEditingController.test.ts
@@ -121,54 +121,19 @@ describe("discrete operations", () => {
   });
 });
 
-describe("paste anchor (#258)", () => {
+describe("paste anchor", () => {
   /**
-   * Reported from the preview: with a placement selected at bar 2 and the
-   * playhead at bar 1, paste landed at bar 1. The playhead is the right anchor
-   * only "with no explicit target selected", which the handler's own comment
-   * says — a live selection *is* that explicit target, and `edit.paste`
-   * declares `ableton: { kind: "follows" }`, where Ableton pastes at the
-   * selection. Verified against Ableton by the reporter.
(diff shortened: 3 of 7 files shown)


afternoon:
## CI: `typecheck, check, unit + component tests` red — #289, not this PR

Failed on [run 33124399255](https://github.com/afternoon/solid-groove/actions/runs/33124399255/job/98698915830), head `4b904e9`. **Every test passed** — `177 passed (177)` files, `2366 passed (2366)` tests. The job exits 1 on one *unhandled* error outside any test:

```
⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯
TypeError: Failed to execute 'dispatchEvent' on 'EventTarget': parameter 1 is not of type 'Event'.
 ❯ Object.exports.convert     node_modules/jsdom/lib/jsdom/living/generated/Event.js:22:9
 ❯ AudioContext.dispatchEvent node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:236:24
 ❯ propagateEvent             node_modules/node-web-audio-api/js/lib/events.js:9:15
 ❯ AudioContext.<anonymous>   node_modules/node-web-audio-api/js/AudioContext.js:102:7
```

This is #289, already filed and root-caused: `installWebAudioTeardownGuard()` installs its wrapper on `nwaa.AudioNode.prototype`, but the throwing dispatch comes from an `AudioContext`, which is not an `AudioNode` — verified against the installed package, `isPrototypeOf(AudioNode.prototype, AudioContext.prototype)` is `false`. `AudioContext.js:102` is the context's own `onstatechange` native callback, firing after jsdom teardown.

Why it is not this PR's:

1. **The diff contains no audio code.** It touches `src/arrangement/` and `src/editor/` only; `src/audio/` is untouched, and `placementEditingController.ts` is byte-identical to `main`.
2. **It is not tied to a test file.** The first occurrence tonight (on #286) reported `src/audio/TrackAudioGraph.mixer.test.ts`; this one reports `src/editor/EditorView.test.tsx`. Different file, identical stack — which is what a teardown race looks like, and is evidence *for* #289's diagnosis rather than against it.
3. **It does not reproduce locally.** `bun run test:all` — the exact command CI runs — was green on this tree: `177 passed`, `2366 passed`, no `Errors` line.

**Action taken:** re-ran the failed job once — the single re-run those rules allow. No test skipped, disabled or quarantined; no empty commit.

**No fix ported, because none exists to port.** #289 has no PR yet. The fix is small and known (extend the guard to the context prototypes, reusing the existing narrow `isStaleEventDispatchTypeError` match), but it belongs in `src/audio/` and pulling it in here would widen a PR that was just deliberately narrowed. It should land as its own change, and this PR will go green on its own once it does — or on a re-run, as it did for #286.

If the re-run comes back red, that changes the reading and I will treat it as this PR's to root-cause.

---

Issue #255, opened 2026-08-07
Horizontal sliders have thumbs and don't follow same design language as vertical sliders


afternoon:
Five raw `<input type="range">` controls never went through `FillSlider`: native thumb, no fill, and two of them dispatch a full command on every pointer move.

- Remove the thumb
- Add the fill
- Make sure updating on drag works (one drag = one history entry, one revision, one save)

## Scope (revised 2026-09-24)

| Surface | Controls | Action |
| --- | --- | --- |
| `src/editor/PianoRollNote.tsx` | Per-note velocity, inside the note block | **Remove the slider.** The note block is too cramped for a usable control, so there is no compact `FillSlider` variant. |
| `src/editor/StepEditor.tsx` | Selected step's velocity | Migrate to `FillSlider` + `createControlGesture` |
| `src/editor/DrumMachinePanel.tsx` | Per-pad pitch, level, pan | Migrate to `FillSlider` + `createControlGesture` |

## Acceptance

- [ ] Piano-roll note blocks carry no velocity slider; dragging, resizing and selecting notes are unaffected, and a note's velocity is preserved.
- [ ] No native thumb on the remaining four controls; the filled portion is the value.
- [ ] Value tracks the pointer live during a drag, on screen and in audio.
- [ ] One drag is one history entry, one revision, one save — asserted, not assumed.
- [ ] Keyboard operation survives on the remaining four.

afternoon:
## Triage against `main` (`2be82d9`) — scope settled with @afternoon

### Which sliders this is about

Not the horizontal `FillSlider`: the mixer's pan control already renders thumbless and fills out from centre (`Mixer.tsx:537`, `orientation="horizontal" bipolar`), and `FillSlider.css:182-193` keeps the thumb transparent in horizontal mode. That one is fine.

The offenders are five raw `<input type="range">` elements that never went through `FillSlider` at all — native horizontal track, native thumb, no fill, no value readout. **All three surfaces are in scope:**

| Surface | Controls |
| --- | --- |
| `src/editor/DrumMachinePanel.tsx:121,135,149` | Per-pad pitch, level, pan |
| `src/editor/StepEditor.tsx:234` | Selected step's velocity |
| `src/editor/PianoRollNote.tsx:43` | Per-note velocity, inside the note block |

[...]

PR #286, opened 2026-08-27
Fix #254 (1/1): Run instrument slider drags as one live gesture

### LINKED ISSUE #254: Slider controls on instruments (sampler) don't update during drag
Dragging instrument slider controls doesn't update during drag, e.g. synth filter cutoff, or sampler pitch.

Mixer slider controls _do_ drag. These should be the same basic component.

Fix is to reuse the mixer slider component as the canonical slider component for all vertical sliders.

afternoon:
## What &amp; why

`SamplerPanel` and `SynthPanel` passed a literal no-op `onInput` and wrote state only in `onCommit`. `FillSlider` is controlled — it paints its fill and its `<output>` readout from `value`, which is read back out of project state — so with nothing applied per tick there was no new state to paint, and the control (and the audio graph reading the same state) sat frozen under the pointer until release.

**The fix suggested in the issue body does not apply and was not done.** It asks to "reuse the mixer slider component as the canonical slider component", but that migration already happened: `SamplerPanel`, `SynthPanel` and `Mixer` all already render the same `FillSlider`. What differed was the *gesture wiring* — the mixer routed every move through a private `createControlGesture` helper, and the instrument panels hand-rolled around it. Re-verified before planning; written up on the issue.

So this extracts that helper and gives the instrument panels the same treatment.

Closes #254

## Second commit: the intermittent lock-up found on the preview

Walking the preview channel turned up a symptom the first commit did not cover: sometimes a slider moved smoothly and the audio followed, sometimes the control **locked up entirely** — no movement, no value change, no message.

Two halves of one defect:

1. **`CommandHistory.beginGesture` throws** when a gesture is already open (`history.ts:210`), and `createControlGesture.input` called it unguarded. So a gesture left open *anywhere* in the editor took down the next control the user touched, from inside its own `input` handler.
2. **Gestures were being left open**, because `FillSlider` only committed on `change` — which never fires when a drag ends somewhere the input does not hear about: released off-element, or the panel unmounted mid-drag by a track switch. One dropped drag then wedged every slider until something else happened to commit.

Both are fixed at their own level. `FillSlider` now commits on pointer up/cancel as well as `change`, closing the gesture at its source — the pattern `PianoRollNote` already uses for its velocity slider. And `createControlGesture` no longer lets a foreign open gesture escape as an exception: it falls back to plain dispatch, so the value still lands and the control still follows the pointer. The cost in that fallback is one history entry per step instead of one per drag — worse to undo, but never a dead control.

Note that half of this was **pre-existing on `main`**, latent in the mixer's private copy of the helper, which is byte-identical. It was rarely hit there (two faders per strip, little unmounting); it became reachable when the same helper started serving seven sliders in a panel that unmounts on every track switch.

## Core flows

Untouched. A bug fix's contract is its regression test, not a core flow. `bun run verify:core-flows` reports 3 flows with one spec each, CF-002/CF-003 still `test.fixme` exactly as on `main` — no flow spec, `docs/core-flows.md`, or `docs/prd.md` was edited, and no `test.fixme` marker was removed.

## Where the helper went

`src/commands/controlGesture.ts`, exported from the `src/commands` barrel. It is a framework-free adapter over `beginGesture`/`dispatch` — no Solid, Firebase or Tone imports, so `commands/boundaries.test.ts` still passes — which puts it beside `history.ts`, the module that owns `beginGesture`. Both `editor/` and `instrument/` already import from `../commands`, so #255's drum-machine, step-editor and piano-roll sliders can reach it without a new dependency direction.

**The move is only a move.** The mixer's private copy is deleted rather than duplicated, and the function body was unchanged in the extracting commit apart from two added doc lines; the lock-up guard arrived as its own later commit, on top. `Mixer.test.tsx` is untouched and green at 23 tests. `beginGesture` is threaded `TrackEditor` → `InstrumentPanel` → both panels; `EditorView` already passed it to `TrackEditor`.

## Two pre-existing tests changed shape — deliberately

`"dispatches an instrument parameter.set once when a slider commits"` in both panel test files now reads the commands the gesture applied instead of the `dispatch` spy. That is not a weakened assertion dodging the change — it is the change: a slider's command sink moves from `dispatch` to `gesture.apply`, which is the whole point of the fix. Both still assert exactly one command with the right scope and `parameterId`.

Every other pre-existing test is byte-identical and passing, including both `"emits nothing per input tick"` tests and the synth `feature_first_use` test.

## Walkthrough

Not captured — the browser suites cannot run in this environment (see Evidence), and `walkthrough:capture` drives the Playwright flow specs.

The change **is** user-visible: sampler pitch / sample start-end / ADSR and synth ADSR / filter cutoff / resonance now move under the pointer — fill, numeric readout and `aria-valuetext` — and the sound changes during the drag instead of jumping on release. Undo is unchanged: one drag is still one step. The regression tests below assert exactly that, including the painted fill extent.

## Evidence

Regression tests, one per panel: `SamplerPanel > follows the pointer mid-drag, and still commits once (#254)` and the matching `SynthPanel` test. Each renders the panel over a real `CommandHistory`, so the slider's `value` really does come back out of the project it edits.

Red first, on the test-only commit with every product file at `origin/main`:

```
 FAIL  |ui| src/instrument/SamplerPanel.test.tsx > SamplerPanel > follows the pointer mid-drag, and still commits once (#254)
TestingLibraryElementError: Unable to find an element with the text: +5 st.

 FAIL  |ui| src/instrument/SynthPanel.test.tsx > SynthPanel > follows the pointer mid-drag, and still commits once (#254)
TestingLibraryElementError: Unable to find an element with the text: 760 Hz.

 Test Files  2 failed (2)
      Tests  4 failed | 9 passed (13)
```

The dumped DOM in that failure is the symptom itself: after dragging pitch to +5 the `<output>` still reads `0 st`, `aria-valuetext="0 st"`, and `<div class="fill-slider-fill" style="height: 50%;">` is unmoved.

For the lock-up, two more tests, each verified red by reverting only its own source change:

- `createControlGesture > still lands values while another gesture is open, instead of throwing` — red with `Error: A gesture is already in progress; commit or cancel it first`.
- `SamplerPanel > commits its gesture on pointer-up, with no change event (#254)` — asserts `history.gestureActive` goes false and exactly one entry lands, with no `change` fired.

| Command | Result |
| --- | --- |
| `bun run typecheck` | pass |
| `bun run check` | pass — biome 503 files, `verify:no-solid-1` clean |
| `bun run test` | pass — 160 files, 2057 tests (run twice, zero failures in the JSON report both times) |
| `bun run verify:test-projects` | pass — one owner per file |
| `bun run verify:core-flows` | pass — 3 flows, one spec each |

**Browser suites did not run.** `bun run test:browser:install` fails to reach the Playwright CDN in this container, which CLAUDE.md anticipates. CI's matrix on push is the cross-browser check for this change; there is no Chromium pre-flight evidence and none should be inferred.

## Known gap, not fixed here

Clicks and pops while dragging were reported on the preview. Not addressed in this PR, and not yet attributed: the sampler cannot be the source (`instruments/sampler.ts:111` `update()` is a plain assignment read on the *next* trigger, so no live node is touched), and the mixer, synth filter and drum pads all ramp over `SMOOTHING_SECONDS`/`MIXER_SMOOTHING_SECONDS` (0.02s). Whether this predates the PR — the mixer has always applied live — is the open question, and it wants its own issue rather than a speculative fix bolted on here.

## Acceptance criteria met

- Instrument sliders track the pointer during a drag, on screen and in audio.
- One drag is still one history entry and one revision — asserted, not assumed (`history.entries` is empty mid-drag, length 1 after `change`, `revision === startRevision + 1`).
- A dropped drag no longer wedges the editor's other controls.
- Scope held to the instrument panels. The drum-machine, step-editor and piano-roll sliders have the mirror-image defect (they track the pointer but write one history entry and one revision *per pointer move*) and are deliberately left for #255, which builds on this same extraction.

### DIFF
--- a/src/commands/controlGesture.test.ts
+++ b/src/commands/controlGesture.test.ts
@@ -1,0 +1,65 @@
+import { describe, expect, it } from "vitest";
+import { createSliceFixtureProject } from "../domain/fixtures";
+import {
+  bareParameterId,
+  readInstrumentParameter,
+  SAMPLER_PITCH,
+} from "../domain/parameters";
+import { createControlGesture } from "./controlGesture";
+import { setParameter } from "./definitions/parameters";
+import { CommandHistory } from "./history";
+
+/**
+ * One drag is one gesture — covered by the panel tests. What is covered here is
+ * the case the panels cannot stage: what a control does when the *rest of the
+ * editor* has left a gesture open. `CommandHistory.beginGesture` throws in that
+ * situation, and a control has no way to know before it asks.
+ */
+describe("createControlGesture", () => {
+  function setup() {
+    const history = new CommandHistory(createSliceFixtureProject());
+    const trackId = history.project.song.tracks[0].id;
+    const pitch = () =>
+      readInstrumentParameter(
+        SAMPLER_PITCH,
+        // biome-ignore lint/suspicious/noExplicitAny: fixture track is a sampler
+        (history.project.song.tracks[0].instrument as any).parameters,
+      );
+    const make = () =>
+      createControlGesture({
+        beginGesture: (options) => history.beginGesture(options),
+        dispatch: (commands) => history.execute(commands),
+        summary: () => "Set pitch",
+        command: (value) =>
+          setParameter(
+            {
+              scope: "instrument",
+              trackId,
+              // The panels address instrument parameters by their bare id.
+              parameterId: bareParameterId(SAMPLER_PITCH.id),
+            },
+            value,
+          ),
+      });
+    return { history, make, pitch };
+  }
+
+  it("still lands values while another gesture is open, instead of throwing", () => {
+    const { history, make, pitch } = setup();
+
+    // A drag that never got its `change` — released off-element, or the panel
+    // unmounted mid-drag — leaves this gesture open.
+    make().input(3);
+    expect(history.gestureActive).toBe(true);
+
+    // The next control the user touches must keep working. Before this, the
+    // `beginGesture` throw escaped through the control's own `input` handler
+    // and the slider locked up: no movement, no value, no message.
+    const next = make();
+    expect(() => next.input(7)).not.toThrow();
+    expect(pitch()).toBe(7);
+
+    next.commit(7);
+    expect(pitch()).toBe(7);
+  });
+});
--- a/src/commands/controlGesture.ts
+++ b/src/commands/controlGesture.ts
@@ -1,0 +1,57 @@
+import type { TransactionResult } from "./execute";
+import type { Gesture, GestureOptions } from "./history";
+import type { RawCommandInput } from "./types";
+
+/**
+ * Drives one continuous control — a fader, a pan, an instrument slider —
+ * through the command layer.
+ *
+ * A drag is one gesture: the first `input` opens it, every later `input`
+ * applies live inside it, and the `change` the browser fires on release closes
+ * it — so the whole drag is one history entry, one revision, one save, and at
+ * most one analytics event, however many pointer moves it took. A keyboard
+ * arrow fires `input` then `change`, so it is one gesture per press.
+ *
+ * Applying every step live is the behaviour, not an optimization: a control
+ * painted from project state sits frozen under the pointer, and the audio
+ * graph reading that state does not move either, if nothing applies until
+ * release (#254). A caller whose `beginGesture` declines still lands each
+ * value, through `dispatch`.
+ */
+export function createControlGesture(props: {
+  beginGesture(options?: GestureOptions): Gesture | undefined;
+  dispatch(
+    commands: RawCommandInput | readonly RawCommandInput[],
+  ): TransactionResult | undefined;
+  /** The history summary for the whole gesture, read when it opens. */
+  summary(): string;
+  /** The command that writes `value`, built fresh for every step. */
+  command(value: number): RawCommandInput;
+}) {
+  let gesture: Gesture | undefined;
+
+  return {
+    input(value: number): void {
+      if (!gesture?.active) {
+        gesture = props.beginGesture({ summary: props.summary() });
+      }
+      const command = props.command(value);
+      if (gesture?.active) {
+        gesture.apply(command);
+      } else {
+        props.dispatch(command);
+      }
+    },
+    commit(value: number): void {
+      // The final `input` already applied this exact value inside the gesture;
+      // closing it is all that is left. With no gesture open (a `change` with
+      // no preceding `input`) the value still has to land.
+      if (gesture?.active) {
+        gesture.commit();
+      } else {
+        props.dispatch(props.command(value));
+      }
+      gesture = undefined;
+    },
+  };
+}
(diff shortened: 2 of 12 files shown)


afternoon:
## CI: one job red, and I do not believe it belongs to this PR

`typecheck, check, unit + component tests` failed on run [33119335942](https://github.com/afternoon/solid-groove/actions/runs/33119335942/job/98681987611). Recording the reasoning rather than waving it through, since "flake" is not a diagnosis.

**Every test passed.** `177 passed (177)` files, `2362 passed (2362)` tests. The job exits 1 on `Errors 1 error` — an *unhandled* error outside any test:

```
❯ Object.exports.convert node_modules/jsdom/lib/jsdom/living/generated/Event.js:22:9
❯ AudioContext.dispatchEvent node_modules/jsdom/lib/jsdom/living/generated/EventTarget.js:236:24
❯ propagateEvent node_modules/node-web-audio-api/js/lib/events.js:9:15
❯ AudioContext.<anonymous> node_modules/node-web-audio-api/js/AudioContext.js:102:7

This error originated in "src/audio/TrackAudioGraph.mixer.test.ts"
```

An `AudioContext` from the `node-web-audio-api` native addon dispatching an event into a jsdom environment that is going away.

Why it reads as not this PR's:

1. **The identical commit passed the identical job** in the other workflow run for the same push — [33119334377 → 98681982571](https://github.com/afternoon/solid-groove/actions/runs/33119334377/job/98681982571), green, alongside the full chromium/firefox/webkit matrix and both emulator suites.
2. **The file is untouched here.** This PR changes `src/commands/controlGesture.ts`, `src/instrument/{FillSlider,SamplerPanel,SynthPanel}.tsx`, `src/editor/{Mixer,InstrumentPanel,TrackEditor}.tsx` and their tests. `src/audio/TrackAudioGraph.mixer.test.ts` and everything under `src/audio/` are untouched — `Mixer.tsx` is the component; `TrackAudioGraph` is the audio graph.
3. **It does not reproduce locally.** `bun run test:all` — the exact command CI runs, not the narrower `bun run test` — three consecutive times: `177 passed`, `2362 passed`, no `Errors` line.
4. **`vitest.config.ts` already documents this class**, at the `audio` project: `node-web-audio-api` is a native addon whose state is shared *per process*, not per worker, so Vitest's module isolation does not reach it. That comment records it measured at ~3 failures in 22 runs before `fileParallelism: false`, and reproducing on `main` before the Solid 2 migration.

**Action taken:** re-ran the failed job once — the single re-run those rules allow, spent here to confirm. No test was skipped, disabled or quarantined, and no empty commit was pushed.

**No fix ported, because none exists yet.** The teardown race looks real but pre-existing and repo-wide, not something this PR introduced or should widen to cover. It likely wants its own issue: `fileParallelism: false` stops two renders overlapping *within* the audio project, but nothing stops a context from outliving its environment at teardown. I have not filed one — say the word and I will, with this trace.

If the re-run comes back red, that changes the reading and I will treat it as this PR's to root-cause.

---

Issue #304, opened 2026-09-22
UI-001 - Three views, a floating dock, and modal sequencing and library


afternoon:
## Scope

The editor stops being a grid of panels and becomes **three views, one job each**, reached from a floating dock:

- **Arrangement** (default) — the timeline, full bleed to the page. No library column, no workspace stack beneath it.
- **Instrument** — one track's instrument, fullscreen and vertical, with a track rail down the left edge for switching. Minimal styling; scale and polish come later.
- **Mixer** — the mixer alone, using the whole page.

Two surfaces become modals over whichever view invoked them:

- **Sequence editing** (step editor / piano roll) — an almost-fullscreen modal over the arrangement, opened by double-clicking a placement. The piano roll in particular gets real estate it has never had.
- **The library** — a modal opened from a sample or preset slot, replacing the always-on left column.

A **dock** floats above the content at the bottom of the app, iOS tab-bar in idiom but in the app's own design language. `1`, `2` and `3` switch views.

## Why

Every DAW puts everything on screen at once and leaves the producer to manage the crowding. This is a deliberate bet the other way: the phone/tablet and hardware idiom — Koala, the Elektron Digitakt, the OP-1 — where you do one job at a time on an uncluttered surface. It may well prove painful, with more switching than a grid layout demands. That is the thing worth learning, and it is cheaper to learn it now than after the arrangement, the device chains and the assistant have all been built into a layout we are not sure about.

PRD principle 4 (*progressive depth* — show the controls needed for the current task) and principle 9 (*opinionated, not restrictive*) are what this is testing in practice.

## Principles for the change

- **Only the UI layer moves.** The domain model, commands, persistence, projections and audio graph should not change. If one of them has to, that is tight coupling the change has exposed, and it is worth reporting on this issue and fixing properly rather than working around.
- **The header stays as it is.** Transport, tempo, position, undo/redo and save status remain at the top; the dock is navigation only. The header's library toggle goes away with the library column.
- **Be fearless with fragile tests.** Tests that only assert the old panel arrangement are deleted, not adapted. A test survives if it asserts something that genuinely cannot be covered at a lower layer — drag and drop being the obvious case.

## Decisions taken up front

| Question | Decision |
| --- | --- |
| Master device chain | Lives in the **Mixer** view, reached by selecting the master strip. Not a fourth view, not a tab. |
[...]

PR #320, opened 2026-09-22
Implement #304 (9/16): Open a clip into the sequence editor, and close it with Escape

afternoon:
## What & why

9/16, builds on #319.

Wires the "open clip" event from 6/16 to mount `SequenceEditor` (7/16) as a modal over `EditorView`. Per the issue's decision table, it is deliberately **not** a `dialog` context — it registers its own `sequence_editor` context (reserved in 1/16) so transport, the note shortcuts, and `1`/`2`/`3` all keep working while it's open, and `Escape` closes it. `useEditorShortcuts.ts` is extended for the new context's dispatch rules. The mock-backend `smoke.spec.ts` gains an `openStarterClip()` helper (which is why 8/16's `data-pixels-per-tick` had to land first).

Directly implements the acceptance criteria for double-click opening the modal, `Escape` closing it, and transport/shortcuts staying live under `sequence_editor` rather than `dialog`.

## Core flows

Untouched directly (CF-008 still needs the library modal and the instrument/mixer views before it can run end to end), but this is the slice that makes the sequence-modal half of CF-008 exercisable for the first time.

## Walkthrough

No UI change asserted by a passing flow yet; covered by `EditorView.test.tsx` ("keeps the transport and the view switches working while it is open" — `3` still switches and `O` still toggles the metronome with the modal open) and the extended `smoke.spec.ts`.

## Evidence

`bun run typecheck`, `bun run test`, `bun run check` pass on this commit. `bun run test:browser:chromium` covers the extended `smoke.spec.ts`.

## Acceptance criteria met

- [x] Double-clicking a placement opens the sequence modal on that clip; `Escape` closes it; transport and the note shortcuts work while it is open, under a `sequence_editor` context rather than `dialog`.

## Safety invariant

`sequence_editor` is a new, additive shortcut context — it changes dispatch only for the modal that is now conditionally mounted, and every existing context's dispatch rule is untouched. Proven by "keeps the transport and the view switches working while it is open," which fails if `sequence_editor` accidentally behaved like `dialog` (which suppresses every other context). `git diff --stat`: 281 insertions / 112 deletions across 4 files, 393 lines by the author's own accounting — under the 400-line cap after the finding-1 split recorded on the issue.

Refs #304

---


afternoon:
**`browser e2e against the emulator (chromium)` is red on `2e0561c`, and it is this PR's.** Root cause found; the fix is below, and I am holding the push for the reason at the end.

**What fails.** Not a test — the suite's *setup project*:

```
1 failed
  [warmup:chromium] › tests/e2e/emulator/warmDevServer.setup.ts:67:1 › warm the dev server's dependency graph
11 did not run
Test timeout of 30000ms exceeded.
```

**Why.** `warmDevServer.setup.ts` walks to the editor and waits on the always-on step editor as its "the editor mounted and `tone` is loaded" sentinel:

```ts
const grid = page.getByRole("region", { name: "Step editor" });
await grid.waitFor({ state: "visible", timeout: 60_000 });
```

This PR is the slice that stops mounting that region on the project page — sequencing moves into the modal. So the sentinel can never appear and the wait hangs.

The file is written so it *cannot* fail ("Why it never fails" in its own header: everything is wrapped in try/catch and it passes regardless). That guarantee is defeated here by a timeout mismatch: the inner waits use 60s and 90s, but the **test-level** timeout is 30s, and a Playwright test timeout aborts the test rather than throwing into the `catch`. So the backstop that was designed to degrade quietly takes the whole suite down instead, and the 11 real tests never run.

**The fix.** Swap the sentinel for something the new shell still mounts on the project page — the arrangement, which is equally proof that the editor mounted and `tone` was pulled in:

```diff
-    const grid = page.getByRole("region", { name: "Step editor" });
-    await grid.waitFor({ state: "visible", timeout: 60_000 });
-    await page.waitForTimeout(2_000);
-    await grid.waitFor({ state: "visible", timeout: 60_000 });
+    const editor = page.getByTestId("arrangement-view-ready");
+    await editor.waitFor({ state: "attached", timeout: 20_000 });
+    await page.waitForTimeout(2_000);
+    await editor.waitFor({ state: "attached", timeout: 20_000 });
```

Worth doing in the same edit: bring those inner timeouts under the 30s test timeout (as above), so the file's "this is an optimization, not an assertion" promise actually holds the next time its sentinel goes stale. That is what turned a warm-up miss into eleven tests that did not run.

This PR already updated `tests/e2e/mock/smoke.spec.ts` for the new shell; the emulator warm-up is the one that got missed. It is a harness file, not a core-flow spec, so it is in scope to edit.

**Why the push is not attached yet.** Every branch from this one up the stack (9/16 through 16/16) carries the same stale sentinel, so this is a stack-wide edit, and the `/implement-feature` workflow still holds these branches while it finishes its walkthrough phase. Pushing now risks a non-fast-forward against its own writes. I will land it across the affected branches as soon as the run reports, and re-run this check then.

---

PR #321, opened 2026-09-22
Implement #304 (10/16): Add the library as a window you open from a slot

afternoon:
## What & why

10/16, builds on #320.

Adds `LibraryModal.tsx`/`.css` and its test suite: the modal surface the library will open into, per the issue's decision "Library access: slot-only. The modal opens from a sample/preset slot." Built and tested standalone here, same pattern as the sequence editor in 7/16 — not yet wired to any slot (that's 11/16).

## Core flows

Untouched — CF-008's library-modal steps and CF-005's rewrite need this mounted and reachable (11/16) first.

## Walkthrough

No UI change — component built but not mounted anywhere reachable yet.

## Evidence

`bun run typecheck`, `bun run test`, `bun run check` pass on this commit. `LibraryModal.test.tsx` is a new, self-contained suite.

## Acceptance criteria met

- None yet directly — the library-modal acceptance criterion (opens from a slot, uses `loadSampleCommands`, closes on insert) is checked once it's wired live in 11/16.

## Safety invariant

Net-new files only (`LibraryModal.tsx`/`.css`/`.test.tsx`), no import from any existing module and no existing file touched. `git diff --stat`: 201 insertions / 0 deletions across 3 files — cannot regress anything that exists today.

Refs #304

---


afternoon:
`browser e2e against the emulator` is red here for the same reason it is red on #320, which this branch builds on: `tests/e2e/emulator/warmDevServer.setup.ts` still waits for the always-on `Step editor` region as its "editor mounted" sentinel, and 9/16 is the slice that stops mounting it. The warm-up's try/catch cannot save it because its inner waits (60s/90s) outlive the 30s test timeout, so the setup project aborts and the real tests never run.

The fix is one edit to that harness file and it has to land on every branch from 9/16 up, including this one. Holding it until the `/implement-feature` run releases these branches, then pushing it across the stack and re-running.

---

PR #322, opened 2026-09-22
Implement #304 (11/16): Open the library from the sampler's slot, and let the arrangement fill the page

afternoon:
## What & why

11/16, builds on #321.

Wires `SamplerPanel`'s sample slot to open `LibraryModal` (10/16), loading a sound through the existing `loadSampleCommands` path and closing the modal on insert — the always-on left library column is removed from `EditorView` in the same slice, since the modal is now the only way in (`EditorHeader`'s now-dead library toggle goes with it, per the issue: "The header's library toggle goes away with the library column"). This is also where the arrangement goes full bleed: with no permanent library column or workspace stack beneath it, `ArrangementView` takes the whole view.

Directly implements the acceptance criteria for the library modal opening from a slot via `loadSampleCommands` and closing on insert, and for the arrangement grid being full bleed.

## Core flows

Untouched directly, but this is what lets CF-008's library steps and the arrangement's full-bleed assertion become checkable once the flow unparks in 16/16.

## Walkthrough

No UI change asserted by a passing flow yet; covered by the rewritten `EditorView.test.tsx` and `SamplerPanel.test.tsx` cases, including "inserts from the keyboard onto the same track a drop would reach," which now also asserts the modal closed.

## Evidence

`bun run typecheck`, `bun run test`, `bun run check` pass on this commit.

## Acceptance criteria met

- [x] The library opens as a modal from a sample or preset slot, loads a sound through the existing `loadSampleCommands` path, and closes on insert.
- [x] The arrangement grid is full bleed: no library column, no workspace beneath it.

## Safety invariant

The insertion path itself is unchanged — `loadSampleCommands` is called exactly as it was before, only the surface presenting it moved from an always-on column to a modal. Proven by "inserts from the keyboard onto the same track a drop would reach" passing with the same command assertions plus the new modal-closed check. `git diff --stat`: 147 insertions / 182 deletions across 9 files (net removal, since the permanent column and header toggle are deleted).

Refs #304

---


afternoon:
`browser e2e against the emulator` is red here for the same reason it is red on #320, which this branch sits above: `tests/e2e/emulator/warmDevServer.setup.ts` still waits for the always-on `Step editor` region, and 9/16 is the slice that stops mounting it. The setup project times out at 30s before its own 60s/90s waits can throw into the try/catch meant to make it harmless, so it aborts and the real tests never run. Browser-independent — chromium fails identically.

One harness edit, needed on every branch from 9/16 up. Holding until the `/implement-feature` run releases these branches, then pushing across the stack and re-running.

---

PR #325, opened 2026-09-22
Implement #304 (14/16): Give the Instrument view a track rail and a home for the device chain

afternoon:
## What & why

14/16, builds on #324.

Adds `EditorInstrument.tsx`/`.css` (the vertical fullscreen Instrument view, replacing the placeholder from 5/16) and `TrackRail.tsx`/`.css` (the left-edge track switcher). Per the issue's decisions table ("Device chains in this issue: layout only"), the view reserves an empty, labelled slot for the device chain that #241 will fill later — it does not build any device-chain UI itself. `TrackInstrument` (split out in 3/16) is what the view mounts per selected track.

Directly implements the Instrument-view acceptance criterion (vertical layout, track rail, reserved chain slot) and stays inside the issue's declared out-of-scope boundary for device chains.

## Core flows

Untouched directly, but this is what lets CF-008's Instrument-view steps (4–5) become checkable once the flow unparks in 16/16.

## Walkthrough

No UI change asserted by a passing flow yet; covered by the new `EditorView.test.tsx` cases ("lists every track down the rail and marks the one on screen").

## Evidence

`bun run typecheck`, `bun run test`, `bun run check` pass on this commit.

## Acceptance criteria met

- [x] The Instrument view shows the selected track's instrument in a vertical fullscreen layout with a track rail down the left edge, and reserves an empty, labelled slot for the device chain.
- [x] Track selection is one piece of state across all three views and the modal (rail selection is proven to drive the shared selection state here; the mixer's side of this is 15/16).

## Safety invariant

The reserved device-chain slot renders an empty, labelled placeholder only — no device-chain logic is introduced, keeping this slice inside the issue's explicit out-of-scope line for #241/#283. Proven by `EditorInstrument.test.tsx`/`EditorView.test.tsx` asserting the slot's label and emptiness. `git diff --stat`: 329 insertions / 43 deletions across 7 files.

This branch passed an Opus review round before this PR was opened.

Refs #304

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KuStRdvh3Q7UfmrwCkbp1H

---
_Generated by [Claude Code](https://claude.ai/code/session_01KuStRdvh3Q7UfmrwCkbp1H)_

PR #326, opened 2026-09-22
Implement #304 (15/16): Give the mixer a master strip, and share the device-chain slot

afternoon:
## What & why

15/16, builds on #325.

Extracts `DeviceChainSlot.tsx`/`.css` — the same empty, labelled reserved-slot component the Instrument view uses (14/16) — and gives `Mixer.tsx` a master strip that, when selected, routes to that shared slot. Per the issue's decisions table ("Master device chain: lives in the Mixer view, reached by selecting the master strip. Not a fourth view, not a tab"), this is layout only, same as 14/16 — #283 fills the slot with real device-chain UI later.

Directly implements the Mixer-view acceptance criterion.

## Core flows

Untouched directly, but this is what lets CF-008's Mixer-view step (and the mixer-fills-the-page assertion) become checkable once the flow unparks in 16/16.

## Walkthrough

No UI change asserted by a passing flow yet; covered by the new `Mixer.test.tsx` case ("opens the master's chain when the master strip is selected").

## Evidence

`bun run typecheck`, `bun run test`, `bun run check` pass on this commit.

## Acceptance criteria met

- [x] The Mixer view fills the page, and selecting the master strip is the route to the master's chain (empty until #283).

## Safety invariant

`DeviceChainSlot` is extracted verbatim from 14/16's inline reserved slot into its own shared component — the Instrument view's rendering is unchanged (a pure extraction, not a rewrite), and the Mixer gains only a new, additive master-strip selection path. Proven by `EditorInstrument.test.tsx` (14/16's tests) continuing to pass unmodified against the extracted component, alongside the new `Mixer.test.tsx` case. `git diff --stat`: 151 insertions / 24 deletions across 7 files.

Refs #304

---


afternoon:
`browser e2e against the emulator` is red here for the same reason it is red on #320, which this branch sits above: `tests/e2e/emulator/warmDevServer.setup.ts` still waits for the always-on `Step editor` region, and 9/16 is the slice that stops mounting it. The setup project times out at 30s before its own 60s/90s waits can throw into the try/catch meant to make it harmless, so it aborts and the real tests never run.

Worth noting where the repair actually is: **16/16 (#327) already carries this exact fix and is green.** It just landed eight slices too late, so every branch from 9/16 to 15/16 — this one included — is red on its own commit. That breaks the property that makes the stack worth stacking (`CLAUDE.md`, "A stack is only worth it if the front can merge without the back": *every slice is green on its own commit*), and as it stands the stack can only be merged all-or-nothing.

The repair is to move that harness hunk down to 9/16 and let 10/16 upward inherit it. Queued for once the `/implement-feature` run releases these branches.

---

